PR #4, opened 2015-06-25
Add ClinicalTrialMapping table to build connection between trial, aleration and tumor type

zhx828:
In order to solve mismatch trial issue, this table has been added.
Previously, if one trial listed under all tumors in one gene(Gene A),
and it ’s also listed under another gene(Gene B) with specific tumor.
All tumors under Gene B will be linked to this trial.


zhx828:
@jjgao  could you please review these changes? Just want to make sure the code looks OK

PR #57, opened 2016-02-10
EvidenceLevel

JiaoJiao123:
bug: return wrong evidence level result when multiple cancer types are
chosen for some mutations

bug cause:
call setLevelOfEvidence function changed original evidence data

bug fix:
clone original evidence item and make change to the copy

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/model/Evidence.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/model/Evidence.java
@@ -172,27 +172,23 @@ public class Evidence implements java.io.Serializable {
         }
         return true;
     }
-
-    public Evidence CloneEvidence(){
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/model/Evidence.java
```diff
@@ -173,7 +173,27 @@ public boolean equals(Object obj) {
         return true;
     }
 
-
+    public Evidence CloneEvidence(){
```
jjgao: Maybe write a copy constructor instead of CloneEvidence. And one small style issue: Java method should start to a lower case letter.

JiaoJiao123:
@jjgao 
JJ I added the copy constructor and tested it in postman. The result looks fine, same with before.

Issue #69, opened 2016-02-26
Auto summary in evidenceCtrl for multiple mutations in same domain region. The oncogenic summary will only be generated once.


zhx828:
In evidenceCtrl, the auto oncogenic summary will be only generated if original evidence doesn't have description. But this could be a issue if there are several mutations have been fill into this region, the first mutation will generate the summary, but the rest mutations will be skipped.

Solution: make a copy of evidence

PR #70, opened 2016-02-26
Create evidence copy when assigning/modifying it

zhx828:
This fixes issue #69


zhx828:
@JiaoJiao123 could you help me do the code review since you created the similar functionality before. Thank you.

PR #63, opened 2016-02-11
Issue #54 annotation missing for braf BRAF-KIAA1549 fusion

JiaoJiao123:
bug: annotation info is missing for some mutations. for example for the query below, only GENE_SUMMARY and GENE_BACKGROUND evidence got returned
{
    "geneStatus": "Complete", 
    "source": "cbioportal", 
    "evidenceTypes": "GENE_SUMMARY,GENE_BACKGROUND,ONCOGENIC,MUTATION_EFFECT,VUS,STANDARD_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_SENSITIVITY,STANDARD_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_RESISTANCE", 
    "queries": [
        
        {
            "hugoSymbol": "BRAF", 
            "alteration": "BRAF-KIAA1549 fusion", 
            "tumorType": "Glioma", 
            "consequence": "fusion", 
            "id": "6342993"
        }
    ]
}

note: Right now, remove the consequence fusion condition check helps to return other evidence type info, but might need to do more test to ensure this is the problem.

Issue #54, opened 2016-02-09
BRAF fusions are not being detected by cBioPortal


debyani-chakravarty:
annotation not showing up for BRAF-KIAA1549 fusio.

https://cbioportal.mskcc.org/beta/case.do?cancer_study_id=mskimpact&case_id=P-0009097

PR #71, opened 2016-02-29
Enable hibernate cache;  Bugs fixed;

zhx828:


zhx828:
Related to #58 #54 #63

PR #264, opened 2016-12-02
Clean up the code

zhx828:


zhx828:
@JiaoJiao123 this is the new pull request

web/yo/app/scripts/controllers/datasummary.js
```diff
@@ -387,110 +387,111 @@ angular.module('oncokbApp')
...
-                        break;
-                    case 'tsv':
-                        downloadOpts.filename += 'tsv';
-                        seperator = '\t';
-                        break;
-                    default:
-                        downloadOpts.filename += 'tsv';
-                        seperator = '\t';
-                        break;
+                switch (type) {
```
JiaoJiao123: there should be an indent for the case statement right?

web/yo/app/scripts/directives/driverealtimestring.js
```diff
@@ -7,91 +7,91 @@
...
-          }, 1000);
-        });
-      },
-      controller: function($scope, $rootScope){
-        $scope.valueChanged = function(newVal) {
-            if(!_.isUndefined($scope.es)){
-                $scope.es.set('vetted', 'uv');
-            }
-        };
+                if (typeof scope.addon !== 'undefined') {
```
JiaoJiao123: same with before, should we use _.isUndefined()? or we could just use if(!scope.addon)? it would capture both undefined and null

web/yo/app/scripts/directives/qtip.js
```diff
@@ -25,17 +25,17 @@ angular.module('oncokbApp')
...
-                                break;
-                            case 'nct':
-                                src += 'https://clinicaltrials.gov/show/' + attrs.number;
-                                break;
-                            case 'abstract':
-                                src += attrs.number;
-                                break;
-                            default:
-                                break;
+                        case 'pmid':
```
JiaoJiao123: indent for case statements

web/yo/app/scripts/controllers/datasummary.js
zhx828: Nope, we no longer use indent for case. You can take a look at this: http://javascript.crockford.com/code.html

web/yo/app/scripts/filters/Filter.js
```diff
@@ -1,127 +1,135 @@
...
-                result.push(object);
+            query = query.toString().toLowerCase();
+            angular.forEach(input, function(object) {
+                if (object.toLowerCase().indexOf(query.toLowerCase()) !== -1) {
+                    result.push(object);
+                }
+            });
+            result.sort(function(a, b) {
+                a = a.toString().toLowerCase();
+                b = b.toString().toLowerCase();
```
JiaoJiao123: Not sure about what data type the result item is. If it is object, should we use JSON.stringify(a)? since a.toString() will return "[object Object]"

Issue #273, opened 2016-12-09
Check the report page. It didn't pick up the prevalence correctly


zhx828:
It didn't pick up the prevalence of colorectal cancer for 16-124

PR #274, opened 2016-12-09
Fix issue #273

zhx828:


web/yo/app/scripts/services/reportitem.js
```diff
@@ -190,13 +190,18 @@ angular.module('oncokbApp')
...
+                        
                         description.forEach(function(e){
                             if(e['Cancer type'].toString().toLowerCase() === 'all tumors' && str.length > 0) {
-                                str.unshift(e.value.toString().trim());
+                                allTumorStr = e.value.toString().trim();
                             }else {
                                 str.push(e.value.toString().trim());
                             }
                         });
+                        if(allTumorStr) {
```
JiaoJiao123: should it be if(allTumorStr.length > 0) ? Since empty string will also be true.

Another thing is, should this statement be moved into the description forEach loop? Since currently, allTumorStr always got overwritten by the previous loop.

Issue #277, opened 2016-12-13
Don't pull other tumor type R1 evidence


zhx828:
R1 is for specific tumor type

PR #283, opened 2016-12-15
Summary updates; Synonymous mutations updates; Allele resistance treatment updates

zhx828:


zhx828:
This closes #276 #277

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -338,9 +338,19 @@ private static String findHighestOncogenic(List<Evidence> evidences) {
     public static List<Alteration> excludeVUS(List<Alteration> alterations) {
         List<Alteration> result = new ArrayList<>();
         Set<Alteration> VUS = new HashSet<>();
-        Set<Gene> allGenes = CacheUtils.getAllGenes();
+        Set<Gene> allGenes = new HashSet<>();
+        if(CacheUtils.isEnabled()) {
+            allGenes = CacheUtils.getAllGenes();
+        }else {
+            allGenes = new HashSet<>(ApplicationContextSingleton.getGeneBo().findAll());
```
zhx828: Nope, this statement will only be used if the cache is not available. In that case, we don't need to set back to cache.

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -455,15 +465,28 @@ private static String findHighestOncogenic(List<Evidence> evidences) {
...
+        // Remove alteration itself
+        alleles.remove(alteration);
+        return filterAllelesBasedOnLocation(alleles, alteration.getProteinStart());
+    }
+
+    public static List<Alteration> getAlleleAndRelevantAlterations(Alteration alteration) {
+        List<Alteration> alleles = getAlleleAlterations(alteration);
+        Alteration oncogenicAllele = AlterationUtils.findOncogenicAllele(alleles);
+
+        if(oncogenicAllele!=null) {
```
zhx828: Yeah, I will format the code.

core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
```diff
@@ -357,6 +357,7 @@
 
         for (Evidence evidence : evidences) {
             Evidence tmpEvidence = new Evidence(evidence);
+            Boolean flag = true;
```
zhx828: No necessary. The memory will be released anyway.

PR #284, opened 2016-12-15
compare oncogenicity

JiaoJiao123:
update oncogenic evidence record only when the oncogenicity level is higher

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/model/Oncogenicity.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/model/Oncogenicity.java
@@ -2,7 +2,10 @@
 
 package org.mskcc.cbio.oncokb.model;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/model/Oncogenicity.java
```diff
@@ -41,5 +44,14 @@ public String getOncogenic() {
     public static Oncogenicity getByLevel(String level) {
         return map.get(level);
     }
+    
+    public static int compareOncogenicity(Oncogenicity o1, Oncogenicity o2){
```
zhx828: Maybe call is compare

core/src/main/java/org/mskcc/cbio/oncokb/model/Oncogenicity.java
```diff
@@ -41,5 +44,14 @@ public String getOncogenic() {
...
+        List<String> oncogenicityValues = new ArrayList<>(Arrays.asList("Inconclusive", "Likely Neutral", "Likely Oncogenic", "Oncogenic"));
+        if(oncogenicityValues.indexOf(o2.getOncogenic()) < oncogenicityValues.indexOf(o1.getOncogenic())){
```
zhx828: return oncogenicityValues.indexOf(o1.getOncogenic()) - oncogenicityValues.indexOf(o2.getOncogenic())

PR #298, opened 2016-12-29
import new structure data

JiaoJiao123:


web/yo/app/scripts/services/stringutils.js
```diff
@@ -1042,12 +1045,18 @@ angular.module('oncokbApp')
...
+                        if (model[e + '_review']) {
+                            object[e + '_review'] = getReview(model[e + '_review']);
+                        }
                     }
                 }
             });
             return object;
         }
-
+        function getReview(model){
```
zhx828: You don't do anything here yet?

PR #319, opened 2017-01-10
Include oncokb.api from property file

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/TumorTypeUtils.java
```diff
@@ -517,4 +517,18 @@ public static OncoTreeType getMappedSpecialTumor(SpecialTumorType specialTumorTy
...
\ No newline at end of file
+
+    private static String getOncoTreeApiUrl() {
+        if (ONCO_TREE_API_URL == null) {
+            try {
+                ONCO_TREE_API_URL = PropertiesUtils.getProperties("oncotree.api");
+            } catch (IOException e) {
+                System.out.print("No oncotree.api specified, will use default setting.");
+            }
+            if (ONCO_TREE_API_URL == null || ONCO_TREE_API_URL.isEmpty()) {
```
JiaoJiao123: Should we use ONCO_TREE_API_URL.trim().isEmpty() ? 
Just in case we have empty spaces as value in the property file

zhx828: Sure

PR #359, opened 2017-02-13
add propagation to the review mode

JiaoJiao123:


core/src/main/java/org/mskcc/cbio/oncokb/model/Evidence.java
```diff
@@ -31,7 +31,6 @@
     private String knownEffect;
     private Date lastEdit;
     private LevelOfEvidence levelOfEvidence;
-    @JsonIgnore
```
zhx828: We shouldn't expose propagation in the API

web/yo/app/scripts/directives/driverealtimestring.js
```diff
@@ -119,13 +119,23 @@ angular.module('oncokbApp')
...
                     scope.changePropagation(true);
-                    scope.$watch('content.propagation', function(n) {
+                    scope.$watch('content.propagation', function(n, o) {
                         if (!_.isUndefined(scope.es)) {
                             if (_.isUndefined(scope.es.get('propagation')) ||
                                 scope.es.get('propagation') !== n) {
                                 scope.es.set('propagation', n);
                             }
                         }
+                        scope.rs.set('lastReviewedPropagation', o);
```
zhx828: If o is null, that means this is the first time to set up the propagation info for the treatment, and usually, that means it's setting up the default value. In this situation, you don't have to trigger the review model.

web/yo/app/scripts/directives/driverealtimestring.js
```diff
@@ -119,13 +119,23 @@ angular.module('oncokbApp')
...
+                        var uuid = scope.uuid.getText();
+                        var tempMapping = $rootScope.reviewMeta.get(uuid);
+                        if(!tempMapping) {
+                            tempMapping = $rootScope.metaModel.createMap();
+                        }
+                        tempMapping.set('review', true);
```
zhx828: Until now, I still don't quite sure about this review property? Does is mean 'to be reviewed', 'reviewed' or something else since it is a Boolean property. We don't have to change for this property this time but in the future, can we avoid this ambiguity naming?

web/yo/app/views/driveRealtimeString.html
```diff
@@ -40,19 +40,27 @@ <h4 ng-if="reviewMode && !rs.get('action')">Old Content:</h4>
     <div ng-if="t === 'treatment-select'" class="treatment-select">
         <div>
             <h4 ng-if="reviewMode && !rs.get('action')">New Content:</h4>
-            <select ng-model="content.stringO" ng-if="!reviewMode || !rs.get('action')"
-                    ng-options="opt.value as opt.label for opt in o">
-            </select>
+            <div ng-if="!reviewMode || !rs.get('action')">
```
zhx828: And what's this `action` again?

web/yo/app/scripts/directives/driverealtimestring.js
JiaoJiao123: when review property is set to true, meaning 'to be reviewed'.

web/yo/app/views/driveRealtimeString.html
JiaoJiao123: originally the action is undefined. we set it during the review process to control the colors and existence of accept and reject icons. Possible values are 'rejected', 'accepted'.

core/src/main/java/org/mskcc/cbio/oncokb/model/Evidence.java
JiaoJiao123: Adding JsonIgnore annotation for propagation will cause not being able to pass propagation from the front end evidence model.

web/yo/app/scripts/directives/driverealtimestring.js
JiaoJiao123: addressed

web/yo/app/views/driveRealtimeString.html
zhx828: Maybe rename this parameter and give me meaningful name, the current name is really confused

JiaoJiao123: "Action" sounds ok for me. Do you have any suggestion? maybe "result"?

PR #364, opened 2017-02-14
display items precisely for review mode

JiaoJiao123:
For example, if oncogenicity is changed, then the whole clinical summary section will be displayed in the review mode, which will cause unnecessary reading work for reviewer.
Therefor, in this PR, we add the check to only display the items where change actually happened.

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -1590,9 +1590,6 @@ angular.module('oncokbApp')
             };
             $scope.getData = function() {
                 var test1 = this.gene;
-                var test = $scope.geneStatus;
-                console.log(test);
-
             };
             $scope.showEntry = function(eStatus) {
                 if (eStatus) {
@@ -1645,7 +1642,7 @@ angular.module('oncokbApp')
             }
 
             $rootScope.reviewMode = false;
-            $scope.displayCheck = function(uuid, reviewObj, mutationReview, tumorReview, treatmentReview) {
+            $scope.displayCheck = function(uuid, reviewObj, mutationReview, tumorReview, treatmentReview, precise) {
                 // regular mode check
                 if (!$rootScope.reviewMode) {
                     if (mutationReview && mutationReview.get('removed') || tumorReview && tumorReview.get('removed') || treatmentReview && treatmentReview.get('removed')) {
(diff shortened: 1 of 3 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -1637,7 +1634,7 @@ angular.module('oncokbApp')
             }
 
             $rootScope.reviewMode = false;
-            $scope.displayCheck = function(uuid, reviewObj, mutationReview, tumorReview, treatmentReview) {
+            $scope.displayCheck = function(uuid, reviewObj, mutationReview, tumorReview, treatmentReview, precise) {
```
zhx828: We probably should add some default settings for these parameters.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1647,10 +1644,19 @@ angular.module('oncokbApp')
                 }
                 // review mode check
                 if (mutationReview && mutationReview.get('removed') || tumorReview && tumorReview.get('removed') || treatmentReview && treatmentReview.get('removed')) {
+                    if(reviewObj && !reviewObj.get('removedItem')) {
```
zhx828: Is this a new status `removedItem`? How many statuses do we have now? Is there a document to explain on each status? We may soon lose the track of all statuses.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: That's a good suggestion. I just added default setting for precise. For the rest, I am not sure if we need to. Since inside the function, we always check if it is null or not before using it.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: That is a very good point. I just put up together [a document ](https://docs.google.com/document/d/1mNMKMOKvigYe9bRqjpmVmJebVbfy_yBus0igKhcj1xI/edit)for all the possible status in _review.

JiaoJiao123:
That is a very good point. I just put up together a document for all the possible status in _review.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1647,10 +1644,21 @@ angular.module('oncokbApp')
...
+                        reviewObj.set('removedItem', true);
+                    }
                     return true;
-                } else if (uuid !== null && checkReview(uuid)) {
-                    return true;
                 }
+                // precisely check for this element
+                precise = _.isBoolean(precise) ? precise : false;
+                if(precise) {
```
zhx828: Should you combine them to `if (_.isBoolean(precise) && precise)`?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1647,10 +1644,21 @@ angular.module('oncokbApp')
...
+                    if (uuid !== null && checkReview(uuid)) {
```
zhx828: Use `return uuid !== null && checkReview(uuid);` directly?

PR #382, opened 2017-02-22
save therapy in NCCN

JiaoJiao123:


zhx828:
Need to be merged in the next release

PR #388, opened 2017-02-22
do not open all for deleted items

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -1657,10 +1657,12 @@ angular.module('oncokbApp')
                         reviewObj.set('removedItem', true);
                     }
                     return true;
+                } else if(reviewObj && reviewObj.get('removedItem')) {
+                    reviewObj.delete('removedItem');
```
JiaoJiao123: yes right. It come from the first commit small fixes.

PR #391, opened 2017-02-23
Include all parent tumor types when querying for relevant tumor types

### LINKED ISSUE #355: Look for all parent nodes when matching tumor type
currently, we only match specific tumor type and main tumor type. We need to look for the parents type and match them.

zhx828:
This will close #355


core/src/main/java/org/mskcc/cbio/oncokb/bo/EvidenceBo.java
```diff
@@ -8,6 +8,7 @@
 import java.util.List;
 
 import org.mskcc.cbio.oncokb.model.*;
+import org.mskcc.oncotree.model.TumorType;
```
JiaoJiao123: can't find oncotree package and also TumorType.java

zhx828: It's maven repository hosted on GitHub OncoTeee repo, mvn-repo branch

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -185,10 +186,10 @@
         return new ResponseEntity<>(sb.toString(), HttpStatus.OK);
     }
 
-    private String getCancerType(OncoTreeType oncoTreeType) {
+    private String getCancerType(TumorType oncoTreeType) {
         return oncoTreeType == null ? null : (
-            oncoTreeType.getSubtype() == null ?
-                oncoTreeType.getCancerType() :
-                oncoTreeType.getSubtype());
+            oncoTreeType.getName() == null ?
```
JiaoJiao123: just wondering, should we use getCode() to check subtype? instead of name?

zhx828: Ideally, both properties shouldn't be empty, but since here we are returning the name, so we should check the name not the code.

PR #397, opened 2017-02-24
Obsolete feature and other bug fixes

JiaoJiao123:


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
```diff
@@ -986,8 +986,10 @@ public static void annotateEvidence(Evidence evidence) throws ParserConfiguratio
                     Article tempAT = articleBo.findArticleByPmid(tempPMID);
                     if(tempAT == null){
                         Article newArticle = NcbiEUtils.readPubmedArticle(tempPMID);
-                        articleBo.save(newArticle);
-                        annotatedArticles.add(newArticle);
+                        if(newArticle != null) {
```
zhx828: Something we should even think about the in the future is that if something failed to be imported (for instance in here, the PMID does not exist), we need to notify the curation teams.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2508,6 +2519,152 @@ angular.module('oncokbApp')
                     }
                 });
             };
+            function obsoletedUUIDs(type, mutation, tumor, TI, treatment) {
```
zhx828: This is the function to get list UUIDs about the obsoleted items? Could you write some comments for this function?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2508,6 +2519,152 @@ angular.module('oncokbApp')
...
+                    return [tumor.prevalence_uuid.getText()];
+                case 'PROGNOSTIC_IMPLICATION':
+                    return [tumor.progImp_uuid.getText()];
+                case 'NCCN_GUIDELINES':
+                    return [tumor.nccn_uuid.getText()];
+                case 'CLINICAL_TRIAL':
+                    return [tumor.trials_uuid.getText()];
+                }
+            };
+            function prepareInsertion(type, mutation, tumor, TI, treatment) {
```
zhx828: Prepare what insertion? Maybe add some comments as well?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3937,13 +4096,14 @@ angular.module('oncokbApp')
             }
 
             function checkWhetherIsDesiredGene() {
```
zhx828: So this function should be renamed?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1866,6 +1865,9 @@ angular.module('oncokbApp')
                 var tempArr = [];
                 for (var i = 0; i < $scope.gene.mutations.length; i++) {
                     var mutation = $scope.gene.mutations.get(i);
+                    if(mutation.name_eStatus.get('obsolete') === 'true') {
```
JiaoJiao123: sure, will do that.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2619,8 +2626,11 @@ angular.module('oncokbApp')
                     }
                 });
             };
+            function checkObsolete(eStatus) {
```
zhx828: This function can be changed to `checkObsolete(object, key)` and use 'name' as default value for the key.  And all `checkObsolete(mutation.name_eStatus)` can be changed to `checkObsolete(mutation)` or `checkObsolete(mutation, 'name')`

Inside the function, use `return object[key_'eStatus'] && object[key + '_eStatus'].get('obsolete') === 'true'`

JiaoJiao123: that's a good point, to make this function more generic. will make the change

Issue #396, opened 2017-02-24
Truncating Mutations is not picked up for `splice_region_variant` without alteration/protein change specified


zhx828:
hugoSymbol: APC
alterationType: splice_region_variant

PR #398, opened 2017-02-25
Fix tumor type summary tests; Bugs fixed

zhx828:
Fix issues #355 #396


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -29,6 +29,16 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
...
+        if (query.getAlteration() == null) {
+            query.setAlteration("");
+        }
+
+        // Need to find a better way to map the alteration name with the variant consequence
+        if (query.getAlteration().isEmpty() && query.getConsequence().equals("splice_region_variant")) {
+            query.setAlteration("splice mutation");
+        }
+
         if (query.getAlteration() != null && query.getAlteration().toLowerCase().matches("gain")) {
```
JiaoJiao123: query.getAlteration() != null checking might not be necessary any more, since you set it to empty when it is null at line 33.

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -321,6 +331,12 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
 
             Collections.sort(sortedEvidence, new Comparator<Evidence>() {
                 public int compare(Evidence e1, Evidence e2) {
+                    Integer comparison = LevelUtils.compareLevel(e1.getLevelOfEvidence(), e2.getLevelOfEvidence());
+
+                    if (comparison != 0) {
+                        return comparison;
+                    }
+
                     if (e1.getId() == null)
```
JiaoJiao123: should we return 0 when both e1.getId() and e2.getId() are null?

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
zhx828: Thanks! I will change that.

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
zhx828: Yeah probably should set to 0

PR #409, opened 2017-03-01
bugFix and exclude empty evidence from saving

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -2211,12 +2215,15 @@ angular.module('oncokbApp')
                     data.knownEffect = 'Resistant';
                     break;
                 case 'CLINICAL_TRIAL':
-                    dataUUID = tumor.trials_uuid.getText();
+                    dataUUID = '';
```
zhx828: You already set it at line 2980

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2106,10 +2106,13 @@ angular.module('oncokbApp')
                     }
                 }
             }
-
+            function getValidUUID(uuid, content) {
```
zhx828: The function name doesn't explain the purpose of this function

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2614,11 +2621,12 @@ angular.module('oncokbApp')
                     typeArr = [ti.name.getText()];
                     dataArr = ti.treatments.asArray();
                     tempType = 'treatment';
+                    unObsoletedEvidences(typeArr, mutation, tumor, ti, null, evidences);
```
zhx828: This name does not explain the purpose of it

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: maybe you mean any other line? since 2980 is an empty line

Issue #416, opened 2017-03-02
evidence/lookup POST is not returning MUTATION_EFFECT


zhx828:
Due to the highestLevel setting

PR #422, opened 2017-03-05
Improve relevant alts mapping

### LINKED ISSUE #401: Map `Truncating Mutations` when query for Deletion if not `Deletion` section curated in that gene

zhx828:
For Fusion, if no `Fusions` has been curated, map `Truncating Mutations`;
For Deletion, if no `Deletion` has been curated, map `Truncating Mutations`;
For Translocation, map `Truncating Mutations`;
For Inversion, map `Truncating Mutations`;
Hanle single gene fusion event(two genes are the same gene);
Map `dup` to `inframe_insertion`;
This pull request will close #401 #416

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
@@ -38,8 +38,8 @@ public class AlterationBoImpl extends GenericBoImpl<Alteration, AlterationDao> i
     public List<Alteration> findMutationsByConsequenceAndPosition(Gene gene, VariantConsequence consequence, int start, int end, List<Alteration> alterations) {
         Set<Alteration> result = new HashSet<>();
 
-        if(gene != null && consequence != null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
```diff
@@ -100,71 +101,100 @@ public Alteration findAlteration(Gene gene, AlterationType alterationType, Strin
...
             }
 
-            //the alteration 'fusions' should be injected into alteration list
-            Alteration alt = findAlteration(alteration.getGene(), alteration.getAlterationType(), "fusions");
-            if (alt != null) {
-                alterations.add(alt);
+            // Map Truncating Mutations to Translocation and Inversion
+            // These two are all structural variants, need a better way to model them.
+            if (alteration.getAlteration().toLowerCase().equals("translocation")
+                || alteration.getAlteration().toLowerCase().equals("inversion")) {
```
JiaoJiao123: should it be .equals("insertion") instead of "inversion"?

PR #443, opened 2017-03-07
update evidence for evidence with multiple cancer types

JiaoJiao123:


web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -222,6 +222,27 @@ ResponseEntity updateVUS(@ApiParam(value = "hugoSymbol", required = true) @PathV
         Set<ClinicalTrial> clinicalTrials = queryEvidence.getClinicalTrials();
         String propagation = queryEvidence.getPropagation();
         
+        List<String> cancerTypes = new ArrayList<>();
+        List<String> subTypes = new ArrayList<>();     
+        Boolean cancerEvidence = true;
```
zhx828: As I mentioned before, can we make the variable name more clear here? `isCancerEvidence` would be easier to understand.

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -222,6 +222,27 @@ ResponseEntity updateVUS(@ApiParam(value = "hugoSymbol", required = true) @PathV
...
+        if(cancerType == null) {
+            cancerEvidence = false;
+        } else {
+            cancerTypes = Arrays.asList(cancerType.split(","));
+            if(subType != null) {
+                subTypes = Arrays.asList(subType.split(","));
+                for(int i = 0;i < subTypes.size();i++) {
```
zhx828: What if the size of subTypes is not the same with cancerTypes?

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -222,6 +222,27 @@ ResponseEntity updateVUS(@ApiParam(value = "hugoSymbol", required = true) @PathV
...
+                    if(subTypes.get(i).equals("null")) {
```
zhx828: If you manually inject the null in the frontend, can we just inject empty string? It will make the request size smaller(very minor)

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -222,6 +222,27 @@ ResponseEntity updateVUS(@ApiParam(value = "hugoSymbol", required = true) @PathV
         Set<ClinicalTrial> clinicalTrials = queryEvidence.getClinicalTrials();
         String propagation = queryEvidence.getPropagation();
         
+        List<String> cancerTypes = new ArrayList<>();
+        List<String> subTypes = new ArrayList<>();     
+        Boolean cancerEvidence = true;
+        if(cancerType == null) {
```
zhx828: I think in most of situation, you pass the empty string instead of null?

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -298,8 +322,64 @@ ResponseEntity updateVUS(@ApiParam(value = "hugoSymbol", required = true) @PathV
                 evidence.setPropagation(propagation);
                 evidenceBo.update(evidence);
             }
-        }
-
+        } else {
+            List<Integer> foundIndices = new ArrayList<>();
```
zhx828: If this is tumor type evidence, you can directly to remove all evidence.

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -231,7 +252,7 @@ ResponseEntity updateVUS(@ApiParam(value = "hugoSymbol", required = true) @PathV
                 knownEffect = oncogenicity.getOncogenic();
             }
         }
-
+        // save newly added evidence    
         if (evidences.isEmpty()) {
```
zhx828: Remember to duplicate oncogenic evidence based on number of mutations in the string

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: EvidenceUtils.annotateEvidence(queryEvidence);
This is called at the begging, which will ensure empty string cancerType get converted to null

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: cancerTypes and subTypes are required to be the same length, I just added a length check

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: That was I did before. But by default, split() removes trailing empty strings from result array, which will mess up subTypes. There are ways to turn that mechanism off, but I prefer to pass in a string "null", just want to make sure subTypes is in the original passed in order.

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: addressed

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: addressed

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
zhx828: ok

PR #445, opened 2017-03-08
bug Fix on getString

JiaoJiao123:
The reason for the break out in getString function is that we use getString() in the getReview function. But inside the review mapping, not all values are string type. For example, currentReviewer is a collaborative string, review/rollback/loading are boolean. 
Since 'currentReviewer', 'review', 'action', 'rollback', 'loading' are only designed for the review process, which don't need to be backed up, we excluded them here.

### DIFF
--- a/web/yo/app/scripts/services/stringutils.js
+++ b/web/yo/app/scripts/services/stringutils.js
@@ -1091,7 +1091,7 @@ angular.module('oncokbApp')
                 if (model.get(e)) {
                     if (model.get(e).type === 'Map') {
                         reviewObj[e] = getReview(model[e]);
-                    } else {
+                    } else if(_.isString(model.get(e))) {
                         reviewObj[e] = getString(model.get(e));
                     }
                 }
@@ -1101,6 +1101,7 @@ angular.module('oncokbApp')
 
         function getString(string) {
             if(!string || !_.isString(string)) {
+                console.log('Value passed in is not a string, please double check');
                 return '';
             }
             var tmp = window.document.createElement('DIV');


web/yo/app/scripts/services/stringutils.js
```diff
@@ -1091,8 +1091,8 @@ angular.module('oncokbApp')
                 if (model.get(e)) {
                     if (model.get(e).type === 'Map') {
                         reviewObj[e] = getReview(model[e]);
-                    } else {
-                        reviewObj[e] = getString(model.get(e));
+                    } else if(['currentReviewer', 'review', 'action', 'rollback', 'loading'].indexOf(e) === -1) {
```
zhx828: This will not prevent similar issue happen in the future, right? Shouldn't we just use _.isString(model.get(e))?

JiaoJiao123:
@zhx828 addressed

web/yo/app/scripts/services/stringutils.js
```diff
@@ -1091,7 +1091,7 @@ angular.module('oncokbApp')
                 if (model.get(e)) {
                     if (model.get(e).type === 'Map') {
                         reviewObj[e] = getReview(model[e]);
-                    } else {
+                    } else if(_.isString(model.get(e))) {
                         reviewObj[e] = getString(model.get(e));
```
zhx828: Another thing, just curious, don't you want to save CollectiveString as well?

JiaoJiao123: what is CollectiveString? 
If you mean collaborative string, then, we don't need to save it so far. Currently only currentReviewer is in collaborative string, which we don't need to save. Since it is only used for review process.

PR #449, opened 2017-03-08
fixing error about the cursor movement

JiaoJiao123:


web/yo/app/scripts/directives/driverealtimestring.js
```diff
@@ -64,7 +64,7 @@ angular.module('oncokbApp')
                         if(scope.rs && _.isNull(scope.rs.get('lastReviewed')) && scope.rs.get('action') !== 'rejected') {
                             scope.rs.set('lastReviewed', o);
                         }
-                        scope.content.stringO = stringUtils.getTextString(scope.object.text);
+                        scope.content.stringO = scope.object.text;
```
zhx828: But this bring back the style inside the text issue, right?

JiaoJiao123: Nope. The cause for the text style issue is I used angualr expression instead of ng-model.

zhx828: Could you take a look this issue ?https://github.com/oncokb/oncokb/commit/3d4f3decb3cbf3f25318d7bf9e72e4ee1162ff70

JiaoJiao123: yeah I made that fix for the text style issue. but That's not right fix. The real fix should be using ng-model for display instead of angular expression

zhx828: Did you make the real fix?

PR #454, opened 2017-03-13
Server side recache other services

### LINKED ISSUE #420: Update ehcache
INFO net.sf.ehcache.util.UpdateChecker - New update(s) found: 2.4.7 [http://www.terracotta.org/confluence/display/release/Release+Notes+Ehcache+Core+2.4]. Please check http://ehcache.org for the latest version.

zhx828:
If we enable second level caching and manually change the database, the
hibernate will not update the cache(change from internal war and
manually trigger the recache even in beta);
This closes #420


zhx828:
@JiaoJiao123 this needs to be tested after review mode is finished. But could you do the code review first? Thanks.

PR #476, opened 2017-03-20
update mutation/tumor/treatment name

JiaoJiao123:


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
```diff
@@ -953,11 +953,35 @@ private static Gene getGene(Integer entrezGeneId, String hugoSymbol) {
...
 
     public static void annotateEvidence(Evidence evidence) throws ParserConfigurationException {
-        ClinicalTrialBo clinicalTrialBo = ApplicationContextSingleton.getClinicalTrialBo();
-        ArticleBo articleBo = ApplicationContextSingleton.getArticleBo();
-        NccnGuidelineBo nccnGuidelineBo = ApplicationContextSingleton.getNccnGuidelineBo();
-        DrugBo drugBo = ApplicationContextSingleton.getDrugBo();
-        TreatmentBo treatmentBo = ApplicationContextSingleton.getTreatmentBo();
+        GeneBo geneBo = ApplicationContextSingleton.getGeneBo();
+        Gene gene = geneBo.findGeneByHugoSymbol(evidence.getGene().getHugoSymbol());
+        evidence.setGene(gene);
```
zhx828: Leave this line after 961

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -276,48 +278,68 @@ private Boolean isEmptyEvidence(Evidence queryEvidence) {
...
+        EvidenceBo evidenceBo = ApplicationContextSingleton.getEvidenceBo();
+        List<Evidence> evidences = evidenceBo.findEvidenceByUUIDs(Collections.singletonList(uuid));
+               
+        EvidenceType evidenceType = queryEvidence.getEvidenceType();
+        // Three special evidences update, which are mutation name change, tumor name change and treatment name change
+        if(evidenceType == null) {
+           // If evidences not exist, return empty list
+           if(evidences.isEmpty())return new ArrayList<Evidence>(); 
+           String specialChangeType = specialTypeCheck(queryEvidence);
+           if(specialChangeType.equals("MUTATION_NAME_CHANGE")) {
```
zhx828: Because specialTypeCheck return null if the queryEvidence not in three special changes, you need to do null check here, otherwise it will throw exception. I just you return empty string instead.

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -256,6 +253,11 @@ private Boolean isEmptyEvidence(Evidence queryEvidence) {
         Set<ClinicalTrial> clinicalTrials = queryEvidence.getClinicalTrials();
         String propagation = queryEvidence.getPropagation();
         
+        // if the gene does not exist, return null
+        if (gene == null) {
+            return null;
```
zhx828: Return empty list. Otherwise you need to do null check on wherever calls this function. Is there any difference in here between returning empty list or null?

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
```diff
@@ -276,48 +278,68 @@ private Boolean isEmptyEvidence(Evidence queryEvidence) {
...
-        
-
+        EvidenceBo evidenceBo = ApplicationContextSingleton.getEvidenceBo();
+        List<Evidence> evidences = evidenceBo.findEvidenceByUUIDs(Collections.singletonList(uuid));
+               
+        EvidenceType evidenceType = queryEvidence.getEvidenceType();
+        // Three special evidences update, which are mutation name change, tumor name change and treatment name change
+        if(evidenceType == null) {
+           // If evidences not exist, return empty list
+           if(evidences.isEmpty())return new ArrayList<Evidence>();
```
zhx828: You can return `evidences`

web/yo/app/scripts/controllers/gene.js
```diff
@@ -437,7 +438,8 @@ angular.module('oncokbApp')
                         if(needReview(tumor.summary_uuid)) {
                             tumorChanged = true;
                         }
-                        if(needReview(tumor.trials_uuid)) {
+                        if(needReview(tumor.trials_uuid) || needReview(tumor.name_uuid)) {
+                            console.log(tumor.cancerTypes_uuid.getText());
```
zhx828: Should we remove all console log if they are no longer needed?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1457,6 +1494,22 @@ angular.module('oncokbApp')
                 return list.join(', ');
             };
 
+            $scope.getOldCancerTypesName = function(cancerTypes) {
```
zhx828: There is a same method under `MainUtils.getCancerTypesName`

web/yo/app/views/gene.html
```diff
@@ -342,7 +364,11 @@ <h3 ng-if="reviewMode">Gene Type:
...
                               <span>{{treatment.type.text}}: </span>
-                              <span ng-click="stopCollopse($event)" ng-class="{editableBox:fileEditable, unEditableBox:!fileEditable}" contenteditable="{{!reviewMode && fileEditable}}" ng-model="treatment.name.text"/></span>
+                              <div ng-if="reviewMode && treatment.name_review.get('lastReviewed')"><br/></div>
+                              <div style="display:inline-block">
+                                   <drive-realtime-string rs="treatment.name_review" uuid="treatment.name_uuid"  es="treatment.name_eStatus" fe="fileEditable" t="'name'" object="treatment.name" ng-click="stopCollopse($event)"></drive-realtime-string>
+                              </div>
+                              <!--<span ng-click="stopCollopse($event)" ng-class="{editableBox:fileEditable, unEditableBox:!fileEditable}" contenteditable="{{fileEditable}}" ng-model="treatment.name.text"/></span>-->
```
zhx828: This is no longer needed?

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: addressed

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: addressed

web/src/main/java/org/mskcc/cbio/oncokb/controller/EvidenceController.java
JiaoJiao123: addressed

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: addressed

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: Yeah right. But I can not use that function since it is getting data from a collaborative list and collaborative string. Here I am getting data from Javascript array and string variables

web/yo/app/views/gene.html
JiaoJiao123: addressed

core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
JiaoJiao123: In the updateEvidenceBasedOnUuid from Evidence Controller, I check if gene is null. Thus we do need to set gene in annotateEvidence even if it is null.

zhx828: Isn't it originally NULL already?

web/yo/app/scripts/controllers/gene.js
zhx828: Maybe combine them? Or at least move it to String utils or Main utils

core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
JiaoJiao123: Hugosymbol is passed in from from end. so the original gene won't be null

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: sure I will put it in the Main Utils

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -85,59 +85,75 @@ angular.module('oncokbApp')
                         return;
```
zhx828: There is a typo on line 23 in the comment

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -232,6 +243,26 @@ angular.module('oncokbApp')
...
+                        case 'MUTATION_NAME_CHANGE':
+                            rejectItem([{obj: $scope.mt.name, reviewObj: $scope.mt.name_review, uuid: $scope.mt.name_uuid}]);
+                            break;
+                        case 'TUMOR_NAME_CHANGE':
+                            var lastReviewed = $rootScope.model.createList();
+                            _.each($scope.tm.cancerTypes_review.get('lastReviewed'), function(ct) {
+                                var cancerType = $rootScope.model.create(OncoKB.CancerType);
+                                cancerType.cancerType.setText(ct.cancerType);
+                                cancerType.subtype.setText(ct.subtype);
+                                cancerType.cancerType_eStatus.set('obsolete', 'false');
```
zhx828: We probably should set obsolete to false in the app.js `OncoKB[_key].prototype.initialize`

web/yo/app/scripts/services/mainutils.js
```diff
@@ -227,6 +227,24 @@ angular.module('oncokbApp')
         }
 
         /**
+         * Output last reviewed cancer type name, either subtype or cancerType
+         * @param {array} cancerTypes array of cancer types
+         * @return {string} TumorType name
+         */
+        function getLastReviewedCancerTypesName(cancerTypes) {
+            var list = [];
+            cancerTypes.forEach(function(cancerType) {
```
zhx828: Please do Array check

web/yo/app/scripts/services/mainutils.js
```diff
@@ -227,6 +227,24 @@ angular.module('oncokbApp')
...
+                if (cancerType.subtype.length > 0) {
```
zhx828: Please do Object and Array check

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -85,59 +85,75 @@ angular.module('oncokbApp')
...
+                            } else {
+                                $scope.modelUpdate($scope.tp, $scope.mt, $scope.tm, $scope.ti, $scope.tt);
+                                $scope.rs.set('action', 'accepted');
+                            }
+                        } else if (dataUUID.length > 0) {
+                            if ($rootScope.isDesiredGene) {
+                                $scope.loading = true;
+                                DatabaseConnector.updateEvidence(dataUUID, data, function(result) {
+                                    if (extraDataUUID.length > 0) {
+                                        DatabaseConnector.updateEvidence(extraDataUUID, extraData, function(result) {
```
JiaoJiao123: Since Clinical Effect is possibly corresponding to two evidences. We send one api call for ONCOGENIC, if there is change for MUTATION_SUMMARY, we send the second api call. But there is better way of doing this. Switch to evidence batch update api is the best option, which is implemented in the code clean PR.

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -232,6 +243,26 @@ angular.module('oncokbApp')
                             $scope.tm.trials.pushAll($scope.tm.trials_review.get('lastReviewed'));
                             rejectItem([{reviewObj: $scope.tm.trials_review, uuid: $scope.tm.trials_uuid}]);
                             break;
+                        case 'MUTATION_NAME_CHANGE':
+                            rejectItem([{obj: $scope.mt.name, reviewObj: $scope.mt.name_review, uuid: $scope.mt.name_uuid}]);
+                            break;
+                        case 'TUMOR_NAME_CHANGE':
+                            var lastReviewed = $rootScope.model.createList();
+                            _.each($scope.tm.cancerTypes_review.get('lastReviewed'), function(ct) {
```
JiaoJiao123: actually not necessary. There is no problem for me to use it this way. Probably because _ is the default reference for underscore.js, which we already included in the index.html

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -232,6 +243,26 @@ angular.module('oncokbApp')
...
+                                var cancerType = $rootScope.model.create(OncoKB.CancerType);
```
JiaoJiao123: It is not necessary. Since app.js is included, and OncoKB is a global variable outside of any angular module, we can use it directly.

web/yo/app/scripts/directives/reviewpanel.js
JiaoJiao123: yeah we could do that but I prefer not to. Since only around 12 eStatus objects are used for the battery. It is unnecessary to initialize them all.

web/yo/app/scripts/directives/reviewpanel.js
JiaoJiao123: addressed

web/yo/app/scripts/directives/reviewpanel.js
JiaoJiao123: addressed

web/yo/app/scripts/directives/reviewpanel.js
JiaoJiao123: addressed

web/yo/app/scripts/services/mainutils.js
JiaoJiao123: addressed

web/yo/app/scripts/services/mainutils.js
JiaoJiao123: addressed

Issue #462, opened 2017-03-16
PMIDs need hover-overs under Tumor Type


debyani-chakravarty:
Please give as much information as possible before submitting your issue - Thank you

Issue: I cannot automatically go to the PMID or read the abstract when I'm in the Prevalence/Prognostic Implications/Standard or Investigational therapeutic implications of sensitivity or resistance.

Example:

Screenshot:
![image](https://cloud.githubusercontent.com/assets/17145213/24006431/21eea010-0a42-11e7-8521-17187ddd2b97.png)

PR #492, opened 2017-03-23
CleanGeneStatus

zhx828:
Instead of indexing by guessing the parent, use indexOf method coming from google drive realtime document

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -1563,21 +1563,17 @@ angular.module('oncokbApp')
 
             function getIndex(mutation, tumor, ti, treatment) {
                 var result = [-1, -1, -1, -1]; // Always return four elements array, standing for mutationIndex, tumorIndex, therapyCategoryIndex and treatmentIndex
-
-                if (mutation && tumor && ti && treatment) {
+                if (mutation) {
                     result[0] = $scope.gene.mutations.indexOf(mutation);
-                    result[1] = mutation.tumors.indexOf(tumor);
-                    result[2] = tumor.TI.indexOf(ti);
-                    result[3] = ti.treatments.indexOf(treatment);
-                } else if (mutation && tumor && ti) {
-                    result[0] = $scope.gene.mutations.indexOf(mutation);
-                    result[1] = mutation.tumors.indexOf(tumor);
-                    result[2] = tumor.TI.indexOf(ti);
-                } else if (mutation && tumor) {
-                    result[0] = $scope.gene.mutations.indexOf(mutation);
-                    result[1] = mutation.tumors.indexOf(tumor);
-                } else if (mutation) {
-                    result[0] = $scope.gene.mutations.indexOf(mutation);
+                    if (tumor) {
+                        result[1] = mutation.tumors.indexOf(tumor);
+                        if (ti) {
+                            result[2] = tumor.TI.indexOf(ti);
+                            if (treatment) {
+                                result[3] = ti.treatments.indexOf(treatment);
+                            }
+                        }
+                    }
                 }
                 return result;
             }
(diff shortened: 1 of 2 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -1538,40 +1538,144 @@ angular.module('oncokbApp')
...
+                        }
+                        return geneStatus[strIndex];
+                    }
+                    ref = loopGetGeneStatusItem(ref, geneStatus[strIndex], indices, ++index, key);
+                }
+                return ref;
+            }
+
+            function loopInitGeneStatus(objects, geneStatus, indices, index) {
+                var defaultIsOpen = Boolean($scope.reviewMode);
```
zhx828: @JiaoJiao123 could you help me specifically check whether I still need to check the review mode here? And also the similar code below.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1538,40 +1538,144 @@ angular.module('oncokbApp')
...
+                    result[0] = $scope.gene.mutations.indexOf(mutation);
+                    result[1] = mutation.tumors.indexOf(tumor);
+                    result[2] = tumor.TI.indexOf(ti);
+                } else if (mutation && tumor) {
+                    result[0] = $scope.gene.mutations.indexOf(mutation);
+                    result[1] = mutation.tumors.indexOf(tumor);
+                } else if (mutation) {
+                    result[0] = $scope.gene.mutations.indexOf(mutation);
+                }
+                return result;
```
JiaoJiao123: It might not be a good idea to duplicate the index searching code that much. maybe we can do something similar like:
if(mutation){
  result[0] = $scope.gene.mutations.indexOf(mutation);
}
if(mutation && tumor) {
  result[1] = mutation.tumors.indexOf(tumor);
}
if(mutation && tumor && ti) {
   result[2] = ..
}
if(mutation && tumor && ti && treatment) {
  result[3] = ...
}

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: no, you don't need to check reviewMode here. You can directly specify var defaultIsOpen = false;

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1538,40 +1538,144 @@ angular.module('oncokbApp')
...
+            function loopInitGeneStatus(objects, geneStatus, indices, index) {
+                var defaultIsOpen = Boolean($scope.reviewMode);
+                if (index < indices.length && indices[index] !== -1) {
+                    var strIndex = indices[index].toString();
+                    if (!geneStatus.hasOwnProperty(strIndex)) {
+                        if (index === 0) {
+                            geneStatus[strIndex] = {
+                                isOpen: defaultIsOpen,
+                                oncogenic: new GeneStatusSingleton(defaultIsOpen),
+                                shortSummary: new GeneStatusSingleton(defaultIsOpen)
```
JiaoJiao123: should we use oncogenic for clinical effect and use effect for mutation effect? It would make sense more. 
And also change correspondingly in the gene.html

web/yo/app/scripts/controllers/gene.js
zhx828: @JiaoJiao123 changed to
```
if (mutation) {
    result[0] = $scope.gene.mutations.indexOf(mutation);
    if (tumor) {
        result[1] = mutation.tumors.indexOf(tumor);
        if (ti) {
            result[2] = tumor.TI.indexOf(ti);
            if (treatment) {
                result[3] = ti.treatments.indexOf(treatment);
            }
        }
    }
}
```

web/yo/app/scripts/controllers/gene.js
zhx828: That may require changing all the data in all documents. It's not a priority.

JiaoJiao123: @zhx828 Isn't geneStatus only used in the gene.js and gene.html? May I know any other documents need to be changed besides gene.html?

zhx828: Dont think so

JiaoJiao123: @zhx828 so could you please change that? Since we are writing new code, It is good to have it make sense.

zhx828:
This will close the issue #462

PR #524, opened 2017-03-30
bug fix on genes page review column

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -2252,29 +2252,39 @@ angular.module('oncokbApp')
...
+                if(!$rootScope.metaData) {
+                    storage.retrieveMeta().then(function(result) {
+                        if (result && result.error) {
+                            dialogs.error('Error', 'Fail to retrieve meta file! Please stop editing and contact the developer!');
                             callback();
-                        });
-                    }
-                });
+                        } else {
+                            storage.getMetaRealtimeDocument(result[0].id).then(function(metaRealtime) {
```
zhx828: `result[0]` will throw exception if there is no meta file in the folder

PR #527, opened 2017-03-31
Bugs fix: Map intragenic; Diff consequence issue

zhx828:
This solves issue https://github.com/oncokb/oncokb/issues/526

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
@@ -100,9 +100,10 @@ public class AlterationBoImpl extends GenericBoImpl<Alteration, AlterationDao> i
         VariantConsequence anyConsequence = VariantConsequenceUtils.findVariantConsequenceByTerm("any");
         alterations.addAll(findMutationsByConsequenceAndPosition(alteration.getGene(), anyConsequence, alteration.getProteinStart(), alteration.getProteinEnd(), fullAlterations));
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
```diff
@@ -178,17 +179,27 @@ public Alteration findAlteration(Gene gene, AlterationType alterationType, Strin
...
                 // If no fusions curated, check the Truncating Mutations.
-                if (truncatingMutation != null) {
+                if (truncatingMutation != null && !alterations.contains(truncatingMutation)) {
+                    alterations.add(truncatingMutation);
+                }
+            }
+
+            // Map intragenic to Deletion or Truncating Mutation
+            if (alteration.getAlteration().toLowerCase().contains("intragenic")) {
+                if (deletion != null) {
```
JiaoJiao123: why not combine two if statement together like what for truncatingMutation below?

zhx828: Because for this case, if deletion is not null, you don't want to include Truncating Mutations.

JiaoJiao123: ok I see. then what about the line 200? I guess the logic there is different?

zhx828: Actually they have same logic. I can merge them together. Give me sec.

PR #533, opened 2017-04-04
highlight text difference, strip css style upon paste and review items by horizontal line

JiaoJiao123:


web/yo/app/index.html
```diff
@@ -170,6 +170,8 @@
     <script src="components/angular-file-upload/angular-file-upload.min.js"></script>
     <script src="components/file-saver.js/FileSaver.js"></script>
     <script src="components/uuid-js/lib/uuid.js"></script>
+    <script src="components/google-diff-match-patch-js/diff_match_patch.js"></script>
+    <script src="components/google-diff-match-patch-js/diff_match_patch_uncompressed.js"></script>
```
zhx828: If I understand correctly from https://github.com/curran/google-diff-match-patch there is not need to include uncompressed version. If the bower config file for this project is actually including both of them, we should change the main file config in bower.json manually.

web/yo/app/scripts/directives/driverealtimestring.js
```diff
@@ -57,8 +57,19 @@ angular.module('oncokbApp')
                 }
                 scope.content.preStringO = scope.content.stringO;
                 $rootScope.$watch('reviewMode', function(n, o) {
-                    scope.reviewMode = n;
+                    if(n !== o) {
+                        scope.reviewMode = n;
+                    }
                 });
+                scope.calculateDiff = function() {
```
zhx828: Very minor, since we don't use this function in the template, can we just make a regular function inside the controller? Just reading a reply from this `http://stackoverflow.com/questions/15018821/should-i-always-bind-functions-to-scope-objects` we should try to make scope as clean as possible?

web/yo/app/index.html
JiaoJiao123: yeah I agree. just did that

PR #548, opened 2017-04-10
Combine clinical effect and biological effect

JiaoJiao123:


core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -608,6 +608,22 @@ private static void setAllAlterations() {
...
     }
 
+    public static Set<Evidence> getEvidencesByUUIDs(Set<String> uuids) {
+        synEvidences();
+
+        Set<Evidence> mappedEvis = new HashSet<>();
+        if (uuids != null) {
+            for (Map.Entry<Integer, Set<Evidence>> map : evidences.entrySet()) {
+                for (Evidence evidence : map.getValue()) {
+                    if (uuids.contains(evidence.getUuid())) {
```
zhx828: Let's also do a null check on evidence as well

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -171,15 +179,37 @@ angular.module('oncokbApp')
...
                                 {obj: $scope.mt.description, reviewObj: $scope.mt.description_review, uuid: $scope.mt.description_uuid}];
+
+                            var oncogenicChange = mainUtils.needReview($scope.mt.oncogenic_uuid);
+                            var effectChange = mainUtils.needReview($scope.mt.effect_uuid) || mainUtils.needReview($scope.mt.description_uuid);
+                            if(oncogenicChange && effectChange) {
+                                $scope.bothChanged = true;
+                                DatabaseConnector.getEvidencesByUUIDs([oncogenicUUID, effectUUID], function(result) {
+                                    var resultJSON = JSON.parse(result.status);
+                                    if (_.isArray(resultJSON)) {
+                                        if(resultJSON[0]) {
```
zhx828: Why not just loop through the resultJSON then?

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -171,15 +179,37 @@ angular.module('oncokbApp')
...
+                                            setUpdateTimeEffectSection(resultJSON[0], oncogenicUUID, effectUUID);
+                                        }
+                                        if(resultJSON[1]) {
+                                            setUpdateTimeEffectSection(resultJSON[1], oncogenicUUID, effectUUID);
+                                        }
+                                    }
+                                    specialCases();
```
zhx828: Why do we go through `specialCases` for ONCOGENIC? I didn't see any switch case inside of that function

JiaoJiao123: Good catch! we don't need to.

PR #560, opened 2017-04-12
Accept all changes by person

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -520,79 +542,87 @@ angular.module('oncokbApp')
                         }
                     }
                 }
-            }
-
+            };
+            $scope.acceptChangesByPerson = function(userName) {
+                if ($scope.status.isDesiredGene) {
```
JiaoJiao123: Nope. If it is not desired gene, we still need to accept changes and update front end models.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -520,79 +546,96 @@ angular.module('oncokbApp')
...
+            };
             function collectChangesByPerson(userName) {
-                $scope.status.savingAll = false;
-                $scope.status.noChanges = false;
-                myUpdatedEvidences = {};
-                myDeletedEvidences = [];
-                myGeneTypeEvidence = {};
-                myUpdatedEvidenceModels = [];
+                if(!userName) {
+                    console.log('Can not collect changes from invalid user name. Please double check the userName.');
```
zhx828: Let's include all username exception into one place. And unify them. In this function, it throws console log, but in previous one, it gives a warning dialog.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -293,7 +294,7 @@ angular.module('oncokbApp')
                         if (reviewObjs[i].get('updateTime')) {
                             mostRecent = i;
                         }
-                    } else if (Date.parse(reviewObjs[mostRecent].get('updateTime')) < Date.parse(reviewObjs[i].get('updateTime'))) {
+                    } else if (reviewObjs[mostRecent].get('updateTime') < reviewObjs[i].get('updateTime')) {
```
zhx828: This worries me. What about the updateTime be the string appears again? Remember the complicated updateTime validate function? Can we reuse them?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -305,8 +306,12 @@ angular.module('oncokbApp')
 
             function setUpdatedSignature(tempArr, reviewObj) {
                 var mostRecent = mostRecentItem(tempArr);
-                reviewObj.set('mostRecentUpdatedBy', tempArr[mostRecent].get('updatedBy'));
-                reviewObj.set('mostRecentUpdateTime', tempArr[mostRecent].get('updateTime'));
+                $scope.realtimeDocument.getModel().beginCompoundOperation();
+                var timeStamp = $scope.realtimeDocument.getModel().create('TimeStamp');
+                timeStamp.value.setText(tempArr[mostRecent].get('updateTime').toString());
```
zhx828: Remember we don't convert the timestamp to string? Is this needed?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -305,8 +306,12 @@ angular.module('oncokbApp')
...
+                timeStamp.by.setText(tempArr[mostRecent].get('updatedBy'));
+                reviewObj.set('mostRecent', timeStamp);
+                $scope.realtimeDocument.getModel().endCompoundOperation();
```
zhx828: Since this is only be set once, we don't have to use begin/end compound

web/yo/app/scripts/controllers/gene.js
```diff
@@ -321,18 +326,15 @@ angular.module('oncokbApp')
...
-                }
-                if(mainUtils.needReview($scope.gene.type_uuid)) {
-                    $scope.status.hasReviewContent = true;
-                    userNames.push($scope.gene.type_review.get('updatedBy'));
-                }
-                if(mainUtils.needReview($scope.gene.background_uuid)) {
-                    $scope.status.hasReviewContent = true;
-                    userNames.push($scope.gene.background_review.get('updatedBy'));
-                }
+                var geneEvisMapping = [{uuid: $scope.gene.summary_uuid, review: $scope.gene.summary_review},
```
zhx828: This works, you can also use `['gene', 'summary', 'background']` and later loop through them and use 
```
if(mainUtils.needReview($scope.gene[key + '_uuid'])) {                    
    $scope.status.hasReviewContent = true;
    userNames.push($scope.gene[key + '_review'].get('updatedBy'));   
}

File: web/yo/app/views/reviewPanel.html [hunk 7]
```diff
@@ -1,19 +1,19 @@
 <div>
     <div class="reviewPanel" ng-if="reviewMode">
         <div ng-if="rt === 'update'">
-            <p class="updateText" ng-if="signatureCheck()">Updated by {{rs.get('mostRecentUpdatedBy') || rs.get('updatedBy')}} at {{(rs.get('mostRecentUpdateTime') || rs.get('updateTime')) | date : 'MMM d, h:mm a yyyy'}}</p>
+            <p class="updateText" ng-if="signatureCheck()">Updated by {{rs.get('mostRecent').by.getText() || rs.get('updatedBy')}} at {{(rs.get('mostRecent').value.getText() || rs.get('updateTime')) | date : 'MMM d, h:mm a yyyy'}}</p>
```
zhx828: For CollaborativeString, you can use get() directly

web/yo/app/scripts/controllers/gene.js
```diff
@@ -599,7 +612,7 @@ angular.module('oncokbApp')
                         evidencesAllUsers[userName].deletedEvidenceModels.push(['mutation', mutation]);
                         continue;
                     }
-                    if (mutation.oncogenic_review.get('mostRecentUpdatedBy') === userName) {
+                    if (mutation.oncogenic_review.get('mostRecent') && mutation.oncogenic_review.get('mostRecent').by.getText() === userName) {
```
zhx828: For CollaborativeString, you can use get() directly

web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -312,9 +312,7 @@ angular.module('oncokbApp')
                                 if(lastReviewedLevel !== null) {
                                     $scope.tt.level.setText(lastReviewedLevel);
                                 }
-                                if(lastReviewedPropagation !== null) {
```
zhx828: We don't need this check anymore?

web/yo/app/views/gene.html
```diff
@@ -365,13 +361,13 @@ <h3 ng-if="reviewMode">Gene Type:
                                  <e-status ng-click="commentClick($event)" ng-keypress="stopCollopse($event)" object="treatment.name_eStatus" id="mutation.name.text + tumor.name.text + ti.name.text + treatment.text" apply-obsolete="applyObsolete(treatment.name_eStatus, 'treatment', mutation, tumor, ti, treatment)"></e-status>
                                 <i ng-click="remove($event, 'treatment', mutation, tumor, ti, treatment)" class="fa fa-trash-o"></i>
                               </span>
-                              <div ng-if="treatment.name_review.get('review') !== 'TREATMENT_NAME_CHANGE'">
+                              <div ng-if="!treNameChangeOnly">
```
zhx828: This variable is a global variable inside of gene.js. Under this scenario, if one treatment name changed for one mutation. All treatments will be shown in the review mode. Right?

JiaoJiao123:
Here is a screenshot of how the button group looks like in the review interface
screen shot 2017-04-14 at 11 00 01 am

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: The reason it was converted to string is to match its type. Because the value in TImeStamp is string type.

zhx828:
@JiaoJiao123 could you bold username?

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: In the current implementation, it is unnecessary for me to do userName check in collectChangesByPerson since it won't be called if userName is invalid. Maybe I should remove it.

web/yo/app/scripts/directives/reviewpanel.js
JiaoJiao123: we shouldn't do the check. If lastReviewedPropagation is null, we should set it back as null.

web/yo/app/views/gene.html
JiaoJiao123: We have displayCheck function for uib-accordion-group to decide if this section need to be displayed or not, treNameChangeOnly only works when there are changes happened in the treatment. But there is a problem rising up here, treNameChangeOnly will be applied to all treatment that has any change. I am working on it.

JiaoJiao123:
After bold the name @zhx828 
screen shot 2017-04-14 at 12 13 43 pm

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: There is no get() listed in the Collaborative method document. I just tried to print out this.gene.name.get(), and it gave me error saying "Uncaught TypeError: this.gene.name.get is not a function"

web/yo/app/scripts/controllers/gene.js
```diff
@@ -290,12 +289,18 @@ angular.module('oncokbApp')
...
-                    if (mostRecent < 0) {
-                        if (reviewObjs[i].get('updateTime')) {
+                    var currentItemTime = new Date(reviewObjs[i].get('updateTime'));
+                    // we only continue to check if current item time is valid
+                    if (currentItemTime instanceof Date && !isNaN(currentItemTime.getTime())) {
+                        if (mostRecent < 0) {
                             mostRecent = i;
+                        } else {
+                            // reset mostRect time when current item time is closer
+                            var mostRecentTime = new Date(reviewObjs[mostRecent].get('updateTime'));
```
zhx828: Can we reuse the code you wrong before for validating the Date? And `mostRecentTime` needs to be checked as well.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -320,24 +323,25 @@ angular.module('oncokbApp')
                 evidencesAllUsers[userName].updatedEvidences = _.extend(evidencesAllUsers[userName].updatedEvidences, evidenceResult);
                 evidencesAllUsers[userName].updatedEvidenceModels.push([type, mutation, tumor, TI, treatment]);
             }
-
+            $scope.getButtonContent = function(x) {
+                return $scope.status[x].savingAll ? 'Saving ' + '<i class="fa fa-spinner fa-spin"></i>' : 'Accept All Changes from <b>' + x + '</b>';
```
zhx828: Please validate the x if you write a function for that.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: mostRecentTime does not need to be checked since we only assign item with the valid time format. And I don't think we can reuse validateTimeFormat function. That function was used to set validated time format and return as string. Here we only need to check if it is valid time.

PR #581, opened 2017-04-25
Improve logic to catch incorrect annotation

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/validation/isoformCheck.java
```diff
@@ -53,6 +51,50 @@ public static void main(String[] args) {
...
+                    if (evidences.size() > 1) {
+                        Set<String> pmids = EvidenceUtils.getPmids(new HashSet<>(evidences));
+                        if (pmids != null && pmids.size() > 0) {
+                            errorMessage.append(" All pmids: " + StringUtils.join(pmids, ", "));
+                        }
+                    }
+                }
+
+                errorMessage.append(" Current ensembl ID (isoform) is " + gene.getCuratedIsoform());
+                if (ensembl.getId().equals(gene.getCuratedIsoform())) {
```
JiaoJiao123: should it be not equal instead of equal?

zhx828: Updated

PR #587, opened 2017-04-26
add curation queue

JiaoJiao123:


web/yo/app/scripts/DataTableSortingMethods.js
```diff
@@ -26,5 +26,17 @@ jQuery.extend(jQuery.fn.dataTableExt.oSort, {
 
     'date-desc': function(a, b) {
         return b - a;
+    },
+    'curated-pre': function(a) {
+        var x = jQuery(a)[0].value;
```
zhx828: No sure whether `jQuery(a)[0]` will be undefined.

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -0,0 +1,142 @@
...
+                        DTColumnDefBuilder.newColumnDef(5).withOption('sType', 'curated'),
+                        DTColumnDefBuilder.newColumnDef(6)
+                    ];
+                },
+                post: function postLink(scope) {
+                    scope.$watch('article', function(n, o) {
+                        if(n && n !== o) {
+                            $timeout.cancel(scope.articleTimeoutPromise);
+                            scope.articleTimeoutPromise = $timeout(function() {
+                                if(/^[\d]*$/.test(scope.article)) {
```
zhx828: `^[\d]*$/` matches empty string as well. Is this ideal?

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -0,0 +1,142 @@
...
+                                    scope.getArticle(scope.article);
+                                }
+                            }, 500);
+                        }
+                    });
+                }
+            },
+            controller: function($scope) {
+                $scope.addCuration = function() {
+                    var item = $rootScope.model.createMap();
```
zhx828: You can initial map in this fashion https://github.com/googledrive/realtime-playground/blob/master/elements/create-document.js#L93

web/yo/app/views/gene.html
```diff
@@ -540,6 +540,11 @@ <h4 ng-if="!tumor.trials_review.get('action') && !tumor.trials_review.get('remov
       </div>
     </div>
   </div>
+    <div style="display: inline-block">
+        <br/><br/>
```
zhx828: Try to use css style instead of line break in here. `<br/>` is supposed to indicate a new line, not for separating sections.

web/yo/app/views/gene.html
```diff
@@ -540,6 +540,11 @@ <h4 ng-if="!tumor.trials_review.get('action') && !tumor.trials_review.get('remov
...
+        <h3>Curation Queue</h3>
+        <curation-queue></curation-queue>
```
zhx828: Why not also include the header into curation-queue?

web/yo/app/views/curationQueue.html
```diff
@@ -0,0 +1,58 @@
...
+            <th>Added by</th>
+            <th>Added at</th>
+            <th>Curator</th>
+            <th>Curated</th>
+            <th>Operation</th>
+        </tr>
+        </thead>
+        <tbody>
+        <tr ng-repeat="x in queue">
+            <td>
```
zhx828: Should we consider separate elements into two sections as a whole? One for x.editable and one for !x.editable. Current way, it works. But highly coupled. I would recommend to separate them.

web/yo/app/scripts/factories/VariantFactory.js
```diff
@@ -264,7 +267,8 @@ angular.module('oncokbApp').factory('DriveAnnotation', ['$http', 'OncoKB', funct
         updateVUS: updateVUS,
         updateEvidenceBatch: updateEvidenceBatch,
         getEvidencesByUUID: getEvidencesByUUID,
-        getEvidencesByUUIDs: getEvidencesByUUIDs
+        getEvidencesByUUIDs: getEvidencesByUUIDs,
+        getpumbedArticle: getpumbedArticle
```
zhx828: Should be called `getPubMedArticle`

web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -321,7 +321,16 @@ angular.module('oncokbApp')
                         fail();
                     });
             }
-
+            function getpumbedArticle(pumbedID, success, fail) {
+                DriveAnnotation
```
zhx828: It should be called pubMedId. Please also validate pubMedId

web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -681,6 +690,7 @@ angular.module('oncokbApp')
                 getSuggestedVariants: getSuggestedVariants,
                 isHotspot: isHotspot,
                 getEvidencesByUUID: getEvidencesByUUID,
-                getEvidencesByUUIDs: getEvidencesByUUIDs
+                getEvidencesByUUIDs: getEvidencesByUUIDs,
+                getpumbedArticle: getpumbedArticle
```
zhx828: Same here, should be named `getPubMedArticle `

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -0,0 +1,142 @@
...
+                    $scope.queue.push({article: item.get('article'), pmid: item.get('pmid'), link: item.get('link'), variant: item.get('variant'), addedBy: item.get('addedBy'), addedAt: item.get('addedAt'), curated: item.get('curated'), curator: item.get('curator')});
+                    $scope.article = '';
+                    $scope.link = '';
+                    $scope.variant = '';
+                    $scope.curator = '';
+                    $scope.predictedArticle = '';
+                    $scope.validPMID = false;
+                };
+                $scope.editCuration = function(index) {
+                    $scope.queue[index].editable = true;
```
zhx828: Ideally, we should validate index. But I don't see any chance this will be undefined if this function is only called in html. But in case we may use this function in the future within js file. Let's still validate the index

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -0,0 +1,142 @@
...
+                };
+                $scope.updateCuration = function(index, x) {
+                    $scope.queue[index].editable = false;
```
zhx828: Same here. We should validate index and queue[index]

web/yo/app/scripts/factories/VariantFactory.js
```diff
@@ -256,6 +256,9 @@ angular.module('oncokbApp').factory('DriveAnnotation', ['$http', 'OncoKB', funct
                 }
             });
     }
+    function getpumbedArticle(pumbedID) {
+        return $http.get('https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json&id=' + pumbedID);
```
zhx828: We should make this function accept array. We can reuse it in other places.

web/yo/app/scripts/DataTableSortingMethods.js
JiaoJiao123: in the current implementation, it will not

web/yo/app/scripts/directives/curationQueue.js
JiaoJiao123: yes that is what we want because we want to reset predicted articles when pmid is empty. but I do need to do some adjustment in getArticle function.

web/yo/app/scripts/directives/curationQueue.js
JiaoJiao123: I don't think it is necessary to validate index, but I do think we need to validate $scope.queue[index]

web/yo/app/scripts/factories/VariantFactory.js
```diff
@@ -256,6 +256,9 @@ angular.module('oncokbApp').factory('DriveAnnotation', ['$http', 'OncoKB', funct
                 }
             });
     }
+    function getPubMedArticle(pubMedIDs) {
+        return $http.get('https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json&id=' + pubMedIDs.join(','));
```
zhx828: Please validate pubMedIds

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1951,6 +1951,28 @@ angular.module('oncokbApp')
                 }
                 dialogs.notify('All Citations', messageContent.join(''), {size: 'lg'});
             };
+            var annotationLocation = {};
```
zhx828: Consider pass `annotationLocation` into `setAnnotationResult`

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -0,0 +1,354 @@
...
+                            }, 500);
+                        }
+                    });
+                }
+            },
+            controller: function($scope) {
+                $scope.allCuration = false;
+                DatabaseConnector.getOncokbInfo(function(oncokbInfo) {
+                    if (oncokbInfo && oncokbInfo.users) {
+                        $scope.curators = oncokbInfo.users;
```
zhx828: Seems this is the first place declare `curators`, you should assign an empty array in `else`  or initialize the array in preLink

PR #593, opened 2017-04-27
Update consequence

zhx828:


zhx828:
This pull request requires to rebuild the database with updated consequence.

PR #596, opened 2017-04-28
collect and save last edit time

JiaoJiao123:
1: collect last edit time in the front end
2: save it in the database and remove MUTATION_SUMMARY

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
@@ -4,6 +4,7 @@
  */
 package org.mskcc.cbio.oncokb.controller;
 
+import org.apache.commons.lang3.StringUtils;
 import org.json.JSONArray;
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -413,17 +404,13 @@ private static void setOncogenic(Gene gene, Alteration alteration, Oncogenicity
                 evidence.setGene(gene);
                 evidence.setAlterations(Collections.singleton(alteration));
                 evidence.setEvidenceType(EvidenceType.ONCOGENIC);
-                if (oncogenic != null) {
```
zhx828: Why removed the if condition

JiaoJiao123: because the oncogenic check is already done in Line 398
if (alteration != null && gene != null && oncogenic != null) {

PR #602, opened 2017-05-04
Add util to check main type exists in cancer types

zhx828:


web/yo/app/scripts/services/mainutils.js
```diff
@@ -227,6 +227,28 @@ angular.module('oncokbApp')
...
+            if (!cancerTypes || !mainType) {
+                return null;
+            }
+            var contain = false;
+            mainType = mainType.toLowerCase().trim();
+            cancerTypes.asArray().forEach(function(cancerType) {
+                // console.log(cancerType.cancerType.text);
+                // console.log(cancerType.cancerType.text.indexOf(mainType));
+                if (cancerType.cancerType && cancerType.cancerType.text.toLowerCase().indexOf(mainType) !== -1) {
+                    contain = true;
```
JiaoJiao123: should we use "return true" instead? There is no need to continue the loop once contain is set to true, right?

web/yo/app/scripts/services/mainutils.js
```diff
@@ -227,6 +227,28 @@ angular.module('oncokbApp')
...
+                }
+            });
+            return contain;
```
JiaoJiao123: return false;

Issue #598, opened 2017-05-01
No row with the given identifier exists: [org.mskcc.cbio.oncokb.model.Alteration#30403] 


zhx828:
Server side returns this error and not returning any info

PR #603, opened 2017-05-04
Turn off query cache

zhx828:
According to this article, query cache should be used with
second_level_cache. Since we implement our own cache, these should be
turned off. Current implementation will run into issue on items
updated(deletion) on other instance, but only refresh our cache
implementation won’t work because the query is still cached in session
http://www.javalobby.org/java/forums/t48846.html `query cache should
always be used in conjunction with the second-level cache.`


zhx828:
This solves #598

PR #623, opened 2017-05-15
Convert other indication treatments' level for quest report

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/quest/VariantAnnotationXML.java
```diff
@@ -322,14 +319,18 @@ private static void exportTherapeuticImplications(Set<TumorType> relevantTumorTy
         //boolean isInvestigational = tagTherapeuticImp.equals("investigational_therapeutic_implications");
         if (!evsSensitivity.get(1).isEmpty() || !evsResisitance.get(1).isEmpty()) {
```
JiaoJiao123: You don't need to do the isEmpty check here right? It will just skip the for loop is the array is empty.
Even if you want to the isEmpty check, it make more sense to put one for loop inside its own empty check.

zhx828: I don't have test cases for this. Don't want to change the original code.

JiaoJiao123: ok I see.

PR #625, opened 2017-05-16
add check for entrez gene id

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -744,7 +744,7 @@ angular.module('oncokbApp')
                     description: null,
                     evidenceType: type,
                     gene: {
-                        entrezGeneId: $scope.meta.gene.entrezGeneId,
+                        entrezGeneId: $scope.meta.gene ? $scope.meta.gene.entrezGeneId : 0,
```
zhx828: `hguoSymbol` should come from $scope.meta.gene as well. So if $scope.meta.gene is not available, this whole gene object shouldn't be contructed. We need to figure out why $scope.meta.gene is not available. 

And second, if entrezGeneId is not available, I wouldn't suggest to use 0. We should remove it.

PR #643, opened 2017-05-25
Update oncogenicity logic and VUS logic

zhx828:


core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
```diff
@@ -31,6 +31,19 @@ public void testProcessQuery() throws Exception {
         indicatorQueryResp = IndicatorUtils.processQuery(query, null, null, null, true);
         assertEquals("The oncogenicity is not matched in variant summary.", "The KRAS V14I mutation is known to be oncogenic.", indicatorQueryResp.getVariantSummary());
 
+        // Check critical case
+        query = new Query(null, null, null, "BRAF", "V600E", null, "Melanoma", null, null, null);
+        indicatorQueryResp = IndicatorUtils.processQuery(query, null, null, null, true);
+        assertTrue("The geneExist in the response should be set to true", indicatorQueryResp.getGeneExist() == true);
```
JiaoJiao123: do you want to replace "indicatorQueryResp.getGeneExist() == true" with "indicatorQueryResp.getGeneExist()"?
Same for other similar assertTrue statement.

JiaoJiao123: do you want to replace "indicatorQueryResp.getGeneExist() == true" with "indicatorQueryResp.getGeneExist()"?
Same for other similar assertTrue statement.

zhx828: I think in this way the logic is relatively more clear. Let's keep in this way for now. I will check some examples to see what convention people usually use.

JiaoJiao123: ok sure.

PR #649, opened 2017-06-06
use bind once annotation

JiaoJiao123:


web/yo/app/views/driveRealtimeString.html
```diff
@@ -100,13 +100,13 @@ <h4 ng-if="reviewMode && reviewObj.get('review') !== false && !reviewObj.get('re
         <div ng-switch-when="select-chosen" class="chosen-full-width">
             <h4 ng-if="reviewMode && reviewObj.get('review') !== false && !reviewObj.get('removedItem') && !reviewObj.get('addedItem') && !reviewObj.get('unobsoleted')">New Content:</h4>
             <select ng-disabled="reviewMode ? (reviewObj.get('review') === false ? true : false) : !fe"
-                    chosen ng-attr-id="{{chosenid}}" allow-single-deselect="true" ng-model="content.stringO"
```
zhx828: Is no longer needed? Are there any styles associated with it?

JiaoJiao123: I didn't see any. This 'select-chosen' is only used for NCCN disease selection. And I didn't see anyplace use its id

Issue #651, opened 2017-06-06
Make Summary/Gene Type/Background not editable for Other biomarkers


zhx828:
Except for Admins.

PR #653, opened 2017-06-08
disable gene editing for other biomarkers, and 2 bug fixes

JiaoJiao123:
1) Releated to Issue #651
2) Bug fixes:
a) clear lastReviewed content for rollback when they enter the review mode, instead of the moment that rollback happens.
b) add userRole check for display obsolete

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -2891,7 +2891,6 @@ angular.module('oncokbApp')
                     // This will only happen if the currentReviewer is not empty
                     $scope.fileEditable = false;
                 }
-                $scope.geneEditable = $scope.fileEditable;
                 // Add timeout until the collaborator join event is triggered.
                 $timeout(function() {
                     if (underOthersReview()) {
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -280,6 +280,9 @@ angular.module('oncokbApp')
             $scope.displayCheck = function(uuid, reviewObj, mutationReview, tumorReview, treatmentReview, precise) {
                 // regular mode check
                 if (!$rootScope.reviewMode) {
+                    if ($scope.gene.name.getText().trim().toLowerCase() === 'other biomarkers' && $scope.userRole !== 8 && !mutationReview) {
+                        $scope.geneEditable = false;
+                    }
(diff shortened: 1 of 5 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -3589,6 +3599,7 @@ angular.module('oncokbApp')
                         'Sorry for any inconvinience.');
                 }
                 $scope.fileEditable = false;
+                $scope.geneEditable = $scope.fileEditable;
```
zhx828: Shouldn't you just make a watcher on fileEditable?

web/yo/app/scripts/directives/estatus.js
```diff
@@ -18,11 +18,6 @@ angular.module('oncokbApp')
             },
             replace: true,
             link: function postLink(scope, element) {
-                $rootScope.$watch('reviewMode', function(n, o) {
```
JiaoJiao123: not in eStatus.js. It was removed in review panel directive js.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: I was thinking about that. Should I? Since we have been trying to reduce watcher usage.

zhx828: We should. Having a watcher is not a bad practice. It's actually designed to deal with the situation like this.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2885,6 +2891,7 @@ angular.module('oncokbApp')
                     // This will only happen if the currentReviewer is not empty
                     $scope.fileEditable = false;
                 }
+                $scope.geneEditable = $scope.fileEditable;
```
zhx828: This can be removed?

JiaoJiao123: I don't think so. This is the first initialization.

zhx828: At line 2886, when the moment you change fileEditable, shouldn't it trigger the watcher to update geneEditable already?

JiaoJiao123: What you said will happen only in some cases. 
When file.editable is false, the watch on fileEditable is not going to be triggered. But at this time, the condition check in displayCheck() will be executed, and geneEditable will be assigned to be false. So I removed it

PR #658, opened 2017-06-13
History

JiaoJiao123:


JiaoJiao123:
@zhx828 Yeah that's a good point. I was aware of that, and plan to do it in another PR. It is hard to test without backup part, I will add it in this PR.

JiaoJiao123:
@zhx828 history backup part is added

JiaoJiao123:
@zhx828 getHistoryData in history backup part is updated. 
When we save JS array as value for collaborative map, it will change the array length to read only. So we can not push items to it until using Array.from() to convert it to a normal array. 
We don't need to use Array.from() when backing up since no need to modify anything.

zhx828:
@JiaoJiao123 thanks! Good to know. The code looks good. But let me do some functional tests before merging

PR #664, opened 2017-06-17
Encode space in URL

zhx828:


zhx828:
This solves issue #661

PR #668, opened 2017-06-23
Include both isoform_mskimpact and isoform_override

zhx828:


zhx828:
@JiaoJiao123 could you take a look? Could you also do full tests to see the newly generated gene document will work properly? Thanks!

web/yo/app/scripts/factories/VariantFactory.js
```diff
@@ -392,8 +392,11 @@ angular.module('oncokbApp').factory('OncoTree', ['$http', 'OncoKB', '_', functio
 angular.module('oncokbApp').factory('ApiUtils', ['$http', function($http) {
     'use strict';
 
-    function getIsoforms() {
-        return $http.get('data/isoformMskcc.json');
+    function getIsoforms(type) {
+        if (type && type === 'msk') {
```
JiaoJiao123: remove check on type exists

web/yo/app/scripts/services/mainutils.js
```diff
@@ -273,10 +277,16 @@ angular.module('oncokbApp')
         function getIsoform(hugoSymbol) {
```
JiaoJiao123: mainutils.js#Line296
using resolve directly without "return"

JiaoJiao123:
@zhx828 Just did the integrated test together with the history PR, it works fine to create a new gene doc, and review mode on that new doc works properly. History model and queue model also work as designed.

PR #675, opened 2017-07-05
bug fix on reject evidences

JiaoJiao123:
1) bug fix on reject Mutation oncogenic or mutation effect evidence
2) validation check on the key of 'lastReviewed'
3) check if new value is the same with the lastReviewed content for clinical trials


web/yo/app/scripts/directives/reviewpanel.js
```diff
@@ -377,8 +377,8 @@ angular.module('oncokbApp')
                                     rejectItem(items);
                                 }
                             } else {
-                                if(oncogenicChange) uuid = oncogenicUUID;
```
zhx828: Why not using `oncogenicUUID` any more?

JiaoJiao123: If you look at line 288 inside the rejectItem function, uuid is the uuid collaborative string object, instead of the real string value.

PR #687, opened 2017-07-27
check review history table

JiaoJiao123:


web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -181,4 +181,49 @@ angular.module('oncokbApp')
                     });
                 }
             };
+            var historyResults;
+            $scope.disableHistoryButton = true;
+            $scope.checkHistoryInputStatus = function() {
+                if (!_.isUndefined($scope.genesForHistory) && $scope.genesForHistory.length > 0) {
```
zhx828: isArray is the proper check

web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -181,4 +181,49 @@ angular.module('oncokbApp')
...
+                $scope.loading = true;
+                if (index === 0) {
+                    historyResults = [];
+                }
+                var documents = Documents.get({title: genesForHistory[index]});
+                var document = _.isArray(documents) && documents.length === 1 ? documents[0] : null;
+                if (document) {
+                    storage.getRealtimeDocument(document.id).then(function(realtime) {
+                        if (realtime && realtime.error) {
+                            console.log('did not get realtime document.');
```
zhx828: Let's deal with the case that realtime doc can not be loaded, maybe show an error message in the same place of the table?

web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -181,4 +181,49 @@ angular.module('oncokbApp')
...
+                        } else {
+                            var model = realtime.getModel();
+                            var historyModel = model.getRoot().get('history');
+                            if (historyModel) {
+                                var historyData = stringUtils.getHistoryData(historyModel).api;
```
zhx828: I assume the api is always an array?

web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -181,4 +181,49 @@ angular.module('oncokbApp')
...
+                                if (historyData.length > 0) {
+                                    _.each(historyData, function(item) {
```
zhx828: You don't have to check empty historyData, _.each will escape empty array anyway

web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -181,4 +181,49 @@ angular.module('oncokbApp')
...
+                                        historyResults.push({gene: genesForHistory[index], admin: item.admin, timeStamp: item.timeStamp, records: item.records});
+                                    });
+                                }
+                            }
+                            if (index === genesForHistory.length - 1) {
+                                $scope.historySearchResults = historyResults;
+                                $scope.historyResultTable = true;
```
zhx828: maybe we should name the var as `showHistoryResultTable`?

web/yo/app/views/reportgenerator.html
```diff
@@ -69,5 +69,65 @@
 
         <br/><br/><br/>
         <loading ng-model="loading" content="Please wait, we are loading gene information."></loading>
+    <hr/>
+    <h3>Review History</h3>
+    <br/>
+    <div class="row">
+        <div class="col-sm-1">Genes:</div>
+        <div class="col-sm-6">
```
zhx828: Why not include Genes: and the dropdown together?

web/yo/app/views/reportgenerator.html
```diff
@@ -69,5 +69,65 @@
...
+        </div>
+        <div class="col-sm-2">
+            <input id="includeUUID" type="checkbox" ng-model="includeUUID"/>
+            <label for="includeUUID">Include UUID</label>
+        </div>
+        <div class="col-sm-2">
+            <button type="button" class="btn btn-default" ng-click="searchHistory(genesForHistory, 0)" ng-disabled="disableHistoryButton"> Submit </button>
+        </div>
+    </div>
+    <br/>
```
zhx828: In general, I'd like to avoid br, use margin/padding instead

web/yo/app/views/reportgenerator.html
JiaoJiao123: to keep them aligned horizontally

web/yo/app/scripts/controllers/reportgenerator.js
JiaoJiao123: yes

web/yo/app/scripts/controllers/reportgenerator.js
JiaoJiao123: I prefer not to do so. The table should only contain history data. If failed to load document, then the user should contact us to fix it.

JiaoJiao123:
Related to Issue #688

PR #689, opened 2017-07-28
Include ActionableGenes, Empty Clinical/Biological Variants into validator

zhx828:
Disable rule 1,2,3,6

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/validation/validation.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/validation/validation.java
@@ -5,6 +5,8 @@
  */
 package org.mskcc.cbio.oncokb.validation;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.services.drive.Drive;
 import com.google.api.services.drive.model.File;
 import com.google.api.services.drive.model.ParentReference;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/validation/validation.java
```diff
@@ -184,38 +214,189 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
...
+        if (value == null)
+            value = "";
+        row.getCustomElements().setValueLocal(element, value);
+    }
+
+    private static String getCancerType(TumorType tumorType) {
+        String cancerTypeName = "";
+        if (tumorType != null) {
+            if (StringUtils.isNullOrEmpty(tumorType.getName())) {
+                if (tumorType.getMainType() != null) {
```
JiaoJiao123: For cancer type name, I understood as we use SubType first. If SubType doesn't exist, then we use MainType. It is not the case here?

core/src/main/java/org/mskcc/cbio/oncokb/validation/validation.java
```diff
@@ -184,38 +214,189 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
...
+
+    private static void printTumorTypeSummary(Set<Evidence> evidences, SpreadsheetService service, WorksheetEntry entry) throws IOException, ServiceException {
+        URL feedUrl = entry.getListFeedUrl();
+        if (evidences != null && service != null && entry != null) {
+            for (Evidence evidence : evidences) {
+                ListEntry row = new ListEntry();
+                setValue(row, "Gene", evidence.getGene().getHugoSymbol());
+
+                List<String> alterationNames = new ArrayList<>();
+                for (Alteration alteration : evidence.getAlterations()) {
```
JiaoJiao123: Put this snippet of getting alterationNames into a function. Since the same code is used in printEvidences()

zhx828: Sure thing

PR #720, opened 2017-09-08
remove R1 Evidence from alternative alles

JiaoJiao123:
Related to Issue #720

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
@@ -235,6 +235,16 @@ public class IndicatorUtils {
 
                     treatmentEvidences = filteredEvis;
                 }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -235,6 +235,16 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
 
                     treatmentEvidences = filteredEvis;
                 }
+                Set<Evidence> evidencesToRemove = new HashSet<>();
+                for (Evidence tempEvidence : treatmentEvidences) {
+                    for (Alteration tempAlteration : tempEvidence.getAlterations()) {
+                        if (tempEvidence.getLevelOfEvidence().equals(LevelOfEvidence.LEVEL_R1) && alleles.contains(tempAlteration)) {
```
zhx828: `tempEvidence.getLevelOfEvidence()` could be null

PR #740, opened 2017-09-25
decompose displayCheck function

JiaoJiao123:


web/yo/app/views/gene.html
```diff
@@ -64,15 +64,15 @@
           </span>
       </div>
 
-    <div ng-if="displayCheck(gene.summary_uuid, gene.summary_review, null, null, null, true)">
+    <div ng-if="displayPrecisely(gene.summary_uuid, gene.summary_review)">
```
zhx828: So these updated place no longer needed all the functionality from displayCheck?

JiaoJiao123: only the other biomarker gene gene editable check is still needed. But it should be put outside of this displayCheck function, because the logics are not related at all. And it will do redundant checks if it stay in displayCheck().

PR #749, opened 2017-10-06
track review status from factory instead of review object

JiaoJiao123:


web/yo/app/scripts/services/mainutils.js
```diff
@@ -454,7 +454,36 @@ angular.module('oncokbApp')
                 return false;
             }
         };
-
+        function isProcessed(type, uuid) {
```
zhx828: Either give a more informative name to this function or move the whole thing to`reviewUtils`. Otherwise, it is confusing when using `MainUtils. isProcessed()`

web/yo/app/scripts/factories/VariantFactory.js
```diff
@@ -446,3 +446,22 @@ angular.module('oncokbApp')
             isHotspot: isHotspot
         };
     }]);
+angular.module('oncokbApp')
+    .factory('ReviewResource', ['$http', 'OncoKB', function() {
+        'use strict';
+        return {
+            reviewMode: false, // reviewMode is tracked in the factory to avoid watchers in directives
```
zhx828: Nice documentation.

PR #770, opened 2017-10-16
Add update priority endpoints; Add priority when parsing whole gene doc

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/model/Treatment.java
```diff
@@ -76,15 +86,16 @@ public void setTreatmentDrugs(Set<TreatmentDrug> treatmentDrugs) {
         }
     }
 
-    public void setDrugs(Set<Drug> drugs) {
+    public void setDrugs(List<Drug> drugs) {
```
JiaoJiao123: why is it necessary to change drugs from Set to List?

PR #777, opened 2017-10-18
get evidences for NAME_CHANE more precisely

JiaoJiao123:
collectUUIDs() provides all of the uuids inside a section, but we only need the evidences uuids. By collecting uuids precicesly, we can save a lot unnecessary checks in the backend.

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -1159,17 +1159,17 @@ angular.module('oncokbApp')
                     }
                     break;
                 case 'MUTATION_NAME_CHANGE':
-                    uuids = collectNameChangesEvidencesUUID('mutation', mutation, []);
+                    uuids = collectUUIDs('mutation', mutation, [], true, true);
                     data.evidenceType = null;
                     historyData.location = mutation.name.getText();
                     break;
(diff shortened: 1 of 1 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -2306,15 +2306,18 @@ angular.module('oncokbApp')
...
-            function collectUUIDs(type, obj, uuids, inside) {
+            function collectUUIDs(type, obj, uuids, inside, evidenceUUIDsOnly) {
                 if (type === 'mutation') {
                     if (!inside) {
                         uuids.push(obj.name_uuid.getText());
                     }
                     uuids.push(obj.oncogenic_uuid.getText());
                     uuids.push(obj.effect_uuid.getText());
-                    uuids.push(obj.description_uuid.getText());
+                    if (!evidenceUUIDsOnly) {
```
zhx828: Just curious if the evidenceUUIDsOnly is null or undefined, then it should behave as false?

JiaoJiao123: @zhx828 Yes, that is correct. This function is used to collect all UUIDs inside a section by default

PR #778, opened 2017-10-18
Update TT summary logcis

zhx828:
Get tumor type specific summary first which includes exact alt, alternative allele, relevant alts then looking for other tumor type summary

Signed-off-by: Hongxin Zhang <jackson.zhang.828@gmail.com>


zhx828:
After exploring some cases, this logic actually does not work.

Issue #781, opened 2017-10-19
Proper handle queue model


zhx828:
I got the issue when the gene model tested without queue model. I was going to just assign an empty array on it, but realize in the same file, there are some other prototype functions call. This needs to be properly handled.
![screen shot 2017-10-18 at 9 06 33 pm](https://user-images.githubusercontent.com/5400599/31749517-78d15b74-b448-11e7-88ad-df47fc332855.png)

PR #803, opened 2017-10-27
move queue to a seperate file and code restrucure based on that

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -3028,33 +3020,11 @@ angular.module('oncokbApp')
...
-                                    $rootScope.metaData = metaRealtime.getModel().getRoot().get('review');
-                                    assignMeta(callback);
-                                }
-                            });
-                        }
+                    additionalFile.load(['all']).then(function(result) {
+                        assignMeta(callback);
+                    }, function(error) {
+                        $scope.fileEditable = false;
+                        callback();
```
zhx828: Usually, we do check on callback whether it is a function. But this function is ok since it is a private function and will only be used in gene.js. But for all functions in utils should be checked.

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -83,41 +72,32 @@ angular.module('oncokbApp')
                         DTColumnDefBuilder.newColumnDef(9),
                         DTColumnDefBuilder.newColumnDef(10)
                     ];
+                    storage.retrieveAllFiles().then(function(result) {
```
zhx828: Just a reminder, all promises need to have fail handler.

web/yo/app/scripts/services/additionalFile.js
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * @ngdoc service
+ * @name oncokb.additionalFile
+ * @description
+ * # Additional Files, currently including meta and queues.
+ * Service in the oncokb.
+ */
+angular.module('oncokbApp')
+    .service('additionalFile', function additionalFile($rootScope, $q, storage, mainUtils, documents, dialogs) {
+        function load(types) {
```
zhx828: Seems like everytime the load function is triggered, we retrieve the file from Google. Should we just keep the meta, and curationQueues alive throughout the application? Basically check whether $rootScope.metaRealtime  and $rootScope.queuesRealtime are available first?

web/yo/app/scripts/services/importer.js
```diff
@@ -73,73 +73,103 @@ angular.module('oncokbApp')
...
         };
+        function backupMeta(parentFolderId) {
+            var deferred = $q.defer();
+            storage.createDocument('Meta', parentFolderId).then(function(file) {
+                console.log('Created meta file');
+                storage.getAdditionalRealtimeDocument(file.id).then(function(metaRealtime) {
+                    if (metaRealtime && metaRealtime.error) {
+                        deferred.reject('Failed to get new meta realtime document');
+                    } else {
+                        var newMetaModel = metaRealtime.getModel();
```
zhx828: I believe line 104-121 can be used to create meta model if the document does not have one. The same thing applies to backupQueues below as well. Probably then this issue can be solved? https://github.com/oncokb/oncokb/issues/781

web/yo/app/scripts/services/additionalFile.js
JiaoJiao123: good point! We should check if it's available first.

web/yo/app/scripts/services/importer.js
JiaoJiao123: In this structure, Meta and Queues documents are mandatory. We can't load the home page if either one is missing. Thus by the point that we can do backuping, we are sure that meta model and queues model do exist, and no need to check. Regarding the issue #781, we only need to worry about the history model, which is solved in this PR by changes in "/databaseconnector.js"

web/yo/app/scripts/services/additionalFile.js
```diff
@@ -0,0 +1,88 @@
...
+            storage.retrieveAdditional().then(function(result) {
+                if (!result || result.error || !_.isArray(result) || result.length !== 2) {
+                    dialogs.error('Error', 'Fail to retrieve additional files! Please stop editing and contact the developer!');
+                    var sendTo = 'dev.oncokb@gmail.com';
+                    var subject = 'Fail to retrieve meta file';
+                    var content = 'The additional files are not correctly located. Please double check. ';
+                    if (result && result.error) {
+                        content += 'System error is ' + JSON.stringify(result.error);
+                    }
+                    mainUtils.sendEmail(sendTo, subject, content);
```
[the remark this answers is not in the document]
JiaoJiao123: we could do that

PR #810, opened 2017-10-31
Support structural variant

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -454,10 +437,95 @@ private static Boolean isVUS(Alteration alteration) {
...
+        List<Alteration> fusionPair = new ArrayList<>();
+        List<Alteration> relevantAlterations = new ArrayList<>();
+        Map<String, Object> map = new HashedMap();
+
+        if (geneStrsList != null) {
+            geneStrsSet = new HashSet<>(geneStrsList);
+        }
+
+        // Deal with two different genes fusion event.
+        if (geneStrsSet.size() >= 2) {
```
JiaoJiao123: would there be cases that the size is more than 2? If so, may I have an example?

PR #814, opened 2017-11-02
Have curation platform suports priority

zhx828:
1. Update priority when moving treatments
2. Add priority when adding new treatment
3. After approving edits on treatment, update priorities

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/bo/GenericBo.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/bo/GenericBo.java
@@ -19,8 +19,6 @@ public interface GenericBo<T> {
 
     void saveOrUpdate(T t);
 
-    void saveOrUpdateAll(List<T> ts);
-
     void delete(T t);
 
     void deleteAll(List<T> ts);
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -1215,17 +1215,22 @@ angular.module('oncokbApp')
                         data.propagation = levelMapping[treatment.name_eStatus.get('propagation')];
                         data.treatments = [];
                         var treatments = treatment.name.text.split(',');
+                        var priorities = getNewPriorities(TI.treatments, [dataUUID]);
                         for (i = 0; i < treatments.length; i++) {
                             var drugs = treatments[i].split('+');
                             var drugList = [];
                             for (var j = 0; j < drugs.length; j++) {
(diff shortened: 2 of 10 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -2892,6 +2907,88 @@ angular.module('oncokbApp')
...
+                    // no need to trigger update.
+                    if (priorities[indexUUid] && priorities[moveIndexUUid]) {
+                        postData[indexUUid] = priorities[indexUUid];
+                        postData[moveIndexUUid] = priorities[moveIndexUUid];
+                    }
+                } else {
+                    postData = priorities;
+                }
+
+                if (Object.keys(postData).length > 0) {
```
JiaoJiao123: needs to do isDesiredGene and internal network check

core/src/main/java/org/mskcc/cbio/oncokb/bo/GenericBo.java
```diff
@@ -19,8 +19,6 @@
 
     void saveOrUpdate(T t);
 
-    void saveOrUpdateAll(List<T> ts);
```
JiaoJiao123: can we do saveOrUpdateAll on Set like this? So we can update a batch
void saveOrUpdateAll(Set<T> ts);

zhx828: As I mentioned in the commit, this method is no longer supported.

web/yo/app/scripts/controllers/gene.js
zhx828: I can add isDesiredGene, but internal check should be inside databaseconnector as wel discussed before.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1758,6 +1763,7 @@ angular.module('oncokbApp')
                     treatment.name_review.delete('added');
                     ReviewResource.accepted = _.union(ReviewResource.accepted, [treatment.name_uuid.getText(), treatment.level_uuid.getText(), treatment.indication_uuid.getText(), treatment.description_uuid.getText()]);
                     clearReview([treatment.name_review, treatment.level_review, treatment.indication_review, treatment.description_review]);
+                    updatePriority(ti.treatments);
```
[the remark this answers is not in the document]
JiaoJiao123: just in case you didn't see it, 'firstLayer' is a parameter already passed into acceptSectionItems().

PR #834, opened 2017-11-08
track failed api calls in meta and make api calls periodically

JiaoJiao123:


web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -376,16 +376,44 @@ angular.module('oncokbApp')
...
             function updateVUS(hugoSymbol, data, success, fail) {
-                DriveAnnotation
-                    .updateVUS(hugoSymbol, data)
-                    .success(function(data) {
-                        success(data);
-                    })
-                    .error(function(error) {
-                        fail(error);
-                    });
+                if ($rootScope.internal) {
```
zhx828: You actually don't have to wrap it if there is nothing will be executed after `setAPIData('vus', hugoSymbol, data);`, just need to do `return;`

web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -376,16 +376,44 @@ angular.module('oncokbApp')
...
+                    if (dataFromFile) {
+                        success('');
+                    } else {
+                        DriveAnnotation
+                            .updateVUS(hugoSymbol, data)
+                            .success(function(data) {
+                                success(data);
+                            })
+                            .error(function(error) {
```
zhx828: If failed, I think we should do `setAPIData ` as well.

web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -376,16 +376,44 @@ angular.module('oncokbApp')
...
+                                var subject = 'VUS update Error for ' + hugoSymbol;
+                                var content = 'The system error returned is ' + JSON.stringify(error);
+                                sendEmail({sendTo: 'dev.oncokb@gmail.com', subject: subject, content: content},
```
zhx828: We have send to dev util.

JiaoJiao123: I would like to use the notifyDeveloper function in MainUtils, but that doesn't work because MainUtils is using DatabaseConnector. If I try to use MainUtils in DatabaseConnector, that will create a Circular Dependency error.

web/yo/app/scripts/services/databaseconnector.js
JiaoJiao123: I am not sure I understand your point. In my view, it is necessary to do the condition check on internal since we only execute setAPIData() when not in the internal network

PR #835, opened 2017-11-08
curation queue and load file bug fixes

JiaoJiao123:
Send the second email notification after queues document gets loaded


JiaoJiao123:
@zhx828 Good catch! But I just realize actually setting notified in both queueItem and queueModelItem is already handled in sendEmail() function. So we don't need to worry about setting thing in secondTimeAutoNotify().

PR #836, opened 2017-11-09
track gene last modified and last saved time stamp in meta

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -1833,7 +1836,7 @@ angular.module('oncokbApp')
 
                 DatabaseConnector.updateGene(params, function(result) {
                     $scope.docStatus.savedGene = true;
-                    changeLastUpdate();
+                    mainUtils.updateLastSavedToDB();
```
zhx828: Just curious, shouldn't these be included in database connector?

web/yo/app/scripts/services/mainutils.js
```diff
@@ -488,6 +488,14 @@ angular.module('oncokbApp')
                 return false;
             }
         }
+        function updateLastModified() {
+            $rootScope.geneTimeStamp.set('lastModifiedBy', user.name);
```
zhx828: I think changes made by developers shouldn't be included?

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: If you are talking about where to call mainUtils.updateLastSavedToDB() after saving changes in Review mode, then it needs to be called in gene.js. Because various APIs are fired depending on what type of update it is, and it is complicated to track all kinds of APIs sent out. But  modelUpdate() in gene.js is always called after APIs calls are finished, that is the correct place to mark last update detail. 
If you are talking about the Update Gene specifically, mainUtils.updateLastSavedToDB() can be called either in gene.js or database connector, it doesn't matter very much. But just to keep consistent with other saving function, I put it in gene.js.

web/yo/app/scripts/services/mainutils.js
JiaoJiao123: If a developer runs a script and changes things systematically, the developers' information won't be written into this timestamp model in meta because this function won't be triggered.
But if a developer edits something manually, then the corresponding information should be marked in timestamp model in meta.

PR #838, opened 2017-11-15
indicate mutation/tumor content info in the headers

JiaoJiao123:


web/yo/app/scripts/controllers/gene.js
```diff
@@ -48,8 +48,25 @@ angular.module('oncokbApp')
                     }
                 }
             };
+            var colorsByLeve = {
```
zhx828: You can leave this part in app.js into rootscope. Similar to what we did for the publci repo

PR #846, opened 2017-11-20
bug fix on the last modified display

JiaoJiao123:
Few genes don't have last modified info on the home page. It turns to be a display bug.


web/yo/app/scripts/controllers/genes.js
```diff
@@ -143,6 +139,19 @@ angular.module('oncokbApp')
                         }
                     }
                 }
+                var genesWithTimeStamp = $rootScope.timeStamp.keys();
+                for (var i = 0; i < genesWithTimeStamp.length; i++) {
+                    var hugoSymbol = genesWithTimeStamp[i];
+                    if (!$scope.metaFlags[hugoSymbol]) {
```
zhx828: You can do $scope.metaFlags[hugoSymbol] = {}; Then reuse the rest of the code.

Issue #821, opened 2017-11-03
Migrating Drug Page to Angular 4


victoria34:
Please give as much information as possible before submitting your issue - Thank you

Issue: 

Example:

Screenshot:

PR #849, opened 2017-11-28
Drug page basically works in Angular 4

victoria34:
Migrate Drugs from AngularJS to Angular. Still need to keep migrating other components like Users to get UserRole.

https://github.com/oncokb/oncokb/issues/821


zhx828:
Base on current plan, we will not upgrade to Angular4

PR #854, opened 2017-12-07
Add EGFR CTD mapping to vIVa/b

zhx828:
This solves issue #853

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
@@ -4,14 +4,12 @@
  */
 package org.mskcc.cbio.oncokb.bo.impl;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
```diff
@@ -114,6 +117,13 @@ public Alteration findAlteration(Gene gene, AlterationType alterationType, Strin
             alterations.add(matchedAlt);
         }
 
+        if (addEGFRCTD(alteration)) {
+            Alteration alt = AlterationUtils.findAlteration(GeneUtils.getGeneByHugoSymbol("EGFR"), "EGFR CTD");
```
JiaoJiao123: From what I see in the curation platform, "EGFR CTD" is corresponding to the name attribute in Alteration model. I doubt the alteration could be found when the cache is not enabled.

Issue #841, opened 2017-11-15
Cache is not updated after updating one evidence


JiaoJiao123:
This is the scenario that I run into the problem. I created a new mutation and then accept it. The first evidence saved will created a new alteration, but the cache is not updated, which cause the second evidence keep creating new mutations.

Issue #839, opened 2017-11-15
entrez_gene_id in Evidence model will be set to 0 if passed in as null


JiaoJiao123:

zhx828:
@JiaoJiao123 entrez_gene_id is required by the database, do you think you can pass in entrez_gene_id when updating the evidence?

JiaoJiao123:
@zhx828 The entrezGeneId is already passed in if the gene meta API runs successfully. 
But there should be an option to query by hugoSymbol only. Could we pass in an invalid entrezGeneId(such as a certain negative number) to indicate that we want to skip query by entrezGeneId?

zhx828:
@JiaoJiao123 I see. I wouldn't suggest that. This should be fixed by me. I can't ask other users do the same thing as well.

PR #855, opened 2017-12-07
Force update alteration cache if new alteration added; Change entrezGeneId type

zhx828:
This solves issue #841 and #839

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
@@ -541,6 +541,10 @@ public class CacheUtils {
         notifyOtherServices("update", entrezGeneId);
     }
 
+    public static void forceUpdateGeneAlterations(Integer entrezGeneId) {
+        alterations.remove(entrezGeneId);
+    }
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -541,6 +541,10 @@ public static void updateGene(Integer entrezGeneId) {
         notifyOtherServices("update", entrezGeneId);
     }
 
+    public static void forceUpdateGeneAlterations(Integer entrezGeneId) {
+        alterations.remove(entrezGeneId);
```
JiaoJiao123: If I understand correctly, this removes the old alterations record for a specific gene. But where do the new alterations got set?

zhx828: When the next time other places try to access this gene's alteration. If CacheUtil finds there is not any alteration available, it will try to recache it from database

JiaoJiao123: oh I see. Got you

PR #860, opened 2017-12-12
Support evidence type in api/search

zhx828:
Not all indicator calls may require all information queried especially when only oncogenic is needed, all therapeutics info any be ignored

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
@@ -20,7 +20,7 @@ public class IndicatorUtils {
         highestLevelOnly = highestLevelOnly == null ? false : highestLevelOnly;
 
         if (evidenceTypes == null || evidenceTypes.isEmpty()) {
-            evidenceTypes.addAll(MainUtils.getAllEvidenceTypes());
+            evidenceTypes = new HashSet<>(MainUtils.getAllEvidenceTypes());
         }
 
         Set<EvidenceType> selectedTreatmentEvidence = new HashSet<>(CollectionUtils.intersection(evidenceTypes, MainUtils.getTreatmentEvidenceTypes()));
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
@@ -14,9 +14,18 @@ import java.util.*;
  */
 public class IndicatorUtils {
     public static IndicatorQueryResp processQuery(Query query, String geneStatus,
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -170,60 +187,64 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
...
+                        indicatorQuery.setOncogenic(oncogenicity.getOncogenic());
+                    }
                 }
 
-                // Only set oncogenicity if no previous data assigned.
-                if (indicatorQuery.getOncogenic() == null && oncogenicity != null) {
-                    indicatorQuery.setOncogenic(oncogenicity.getOncogenic());
+                if (hasTreatmentEvidence) {
+                    treatmentEvidences = EvidenceUtils.keepHighestLevelForSameTreatments(
+                        EvidenceUtils.getRelevantEvidences(query, source, geneStatus,
```
JiaoJiao123: Should we do the null check on the getRelevantEvidences() return before calling keepHighestLevelForSameTreatments()?

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -234,17 +255,19 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
...
-                        EvidenceUtils.convertEvidenceLevel(
-                            EvidenceUtils.getEvidence(Collections.singletonList(oncogenicMutation),
-                                MainUtils.getTreatmentEvidenceTypes(),
-                                (levels != null ?
-                                    new HashSet<>(CollectionUtils.intersection(levels,
-                                        LevelUtils.getPublicAndOtherIndicationLevels())) : LevelUtils.getPublicAndOtherIndicationLevels())), new HashSet<>(oncoTreeTypes))));
+                    if (hasTreatmentEvidence) {
+                        treatmentEvidences.addAll(EvidenceUtils.keepHighestLevelForSameTreatments(
+                            EvidenceUtils.convertEvidenceLevel(
+                                EvidenceUtils.getEvidence(Collections.singletonList(oncogenicMutation),
```
JiaoJiao123: Same questions here. Should we do the null check before passing the return as the parameter for the next round call?

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
zhx828: getRelevantEvidences will always return an empty list if nothing has been found.

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
zhx828: same here, an empty list will always be returned.

Issue #258, opened 2016-12-02
summary API doesn't accept comma in tumorType


zhx828:
Example: PTCH1	H634Kfs*59	Skin Cancer, Non-Melanoma

PR #863, opened 2017-12-13
Drop multi query supports on summary.json GET

zhx828:
This solves https://github.com/oncokb/oncokb/issues/258


JiaoJiao123:
@zhx828 Did you think about changing the separator from "," to a more unique one such as ";"? Just feel like the feature to support multiple tumor types summary query is very nice

zhx828:
@JiaoJiao123 there is no use case like that anymore. Multi-queries should use POST

PR #874, opened 2017-12-26
portal alteration importer mapping fix

JiaoJiao123:
There are two places fixed.
a) proteinPosStart should be used instead of startPosition. Same thing for the endPosition. This caused the portal alterations failed to fall into the oncokb alteration protein range.
b) use entrezGeneId or hugoSymbol to compare gene instead of object


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
```diff
@@ -60,7 +60,8 @@ public Alteration findAlteration(Gene gene, AlterationType alterationType, Strin
         if (gene != null && consequence != null && !consequence.getTerm().equals("NA")) {
             if (alterations != null && alterations.size() > 0) {
                 for (Alteration alteration : alterations) {
-                    if (alteration.getGene().equals(gene) && alteration.getConsequence() != null && alteration.getConsequence().equals(consequence)) {
+                    if ((alteration.getGene().getEntrezGeneId() != null && alteration.getGene().getEntrezGeneId().equals(gene.getEntrezGeneId()) || alteration.getGene().getHugoSymbol() != null && alteration.getGene().getHugoSymbol().equals(gene.getHugoSymbol()))
```
zhx828: Why changed the logic?

JiaoJiao123:
@zhx828 In the PortalAlterationImporter.java, Line 120 and Line 121 are modified. The rest are unchanged.

core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
JiaoJiao123: The original logic was "alteration.getGene().equals(gene)" and it kept returning false. This is the problem I asked you earlier. As I understand, both gene and alteration came from the same cache,  I couldn't figure out why they are different. Sorry it is bad practice to change things I don't understand to just make it work. I will ask you more about this issue tomorrow.

zhx828:
This solves issue https://github.com/oncokb/oncokb/issues/878

Issue #888, opened 2018-01-11
Null Pointer exception for CD74-CD74, structural_variant, Bladder Urothelial Carcinoma DELETION


zhx828:
Please give as much information as possible before submitting your issue - Thank you

Issue: 

Example:

Screenshot:

PR #890, opened 2018-01-12
If gene is not available, return null when finding alteration

zhx828:
This solves #888

### DIFF
--- a/core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
+++ b/core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
@@ -26,6 +26,18 @@ public class IndicatorUtilsTest {
         assertEquals("The oncogenicity is not empty, but it should.", "", indicatorQueryResp.getOncogenic());
         assertTrue("There is treatment(s) in the response, but it should no have any.", indicatorQueryResp.getTreatments().size() == 0);
 
+        query = new Query(null, null, null, "CD74-CD74", null, "structural_variant", StructuralVariantType.DELETION, "Pancreatic Adenocarcinoma", null, null, null, null);
+        indicatorQueryResp = IndicatorUtils.processQuery(query, null, null, null, true);
+        assertEquals("Gene should not exist, but it does.", false, indicatorQueryResp.getGeneExist());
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
```diff
@@ -26,6 +26,18 @@ public void testProcessQuery() throws Exception {
         assertEquals("The oncogenicity is not empty, but it should.", "", indicatorQueryResp.getOncogenic());
         assertTrue("There is treatment(s) in the response, but it should no have any.", indicatorQueryResp.getTreatments().size() == 0);
 
+        query = new Query(null, null, null, "CD74-CD74", null, "structural_variant", StructuralVariantType.DELETION, "Pancreatic Adenocarcinoma", null, null, null, null);
```
JiaoJiao123: Looking at the issue associated with this PR, I wonder should you test "Bladder Urothelial Carcinoma" instead of "Pancreatic Adenocarcinoma"?

core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
```diff
@@ -26,6 +26,18 @@ public void testProcessQuery() throws Exception {
...
+        indicatorQueryResp = IndicatorUtils.processQuery(query, null, null, null, true);
+        assertEquals("Gene should not exist, but it does.", false, indicatorQueryResp.getGeneExist());
+        assertEquals("The oncogenicity is not empty, but it should.", "", indicatorQueryResp.getOncogenic());
+        assertTrue("There is treatment(s) in the response, but it should no have any.", indicatorQueryResp.getTreatments().size() == 0);
+
+        query = new Query(null, null, null, "CD74-CD74", null, "structural_variant", StructuralVariantType.DELETION, "Pancreatic Adenocarcinoma", "fusion", null, null, null);
```
JiaoJiao123: same as above

core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
zhx828: Tumor type does not matter in this case

PR #908, opened 2018-01-24
Support more range of EGFR vIV and also EGFR CTD query

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
```diff
@@ -131,6 +131,13 @@ public Alteration findAlteration(Gene gene, AlterationType alterationType, Strin
             }
         }
 
+        if (alteration.getGene().getHugoSymbol().equals("EGFR") && alteration.getAlteration().equals("CTD")) {
```
JiaoJiao123: Is the alteration should be equal to "EGFR CTD" instead of "CTD"?

JiaoJiao123: What is CTD by the way?

zhx828:
It represents C terminal deletion. The comment you pointed above is to
prevent user searches for CTD. Since I’m the system we curate as EGFR CTD,
not CTD

PR #910, opened 2018-01-25
Tweaks to fix rc branch while saving data

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -529,23 +511,14 @@ private static void synAlterations() {
...
-        genesToBeUpdated.add(entrezGeneId);
-        notifyOtherServices("update", entrezGeneId);
-    }
-
     public static void forceUpdateGeneAlterations(Integer entrezGeneId) {
         alterations.remove(entrezGeneId);
     }
 
-    public static void updateGene(Integer entrezGeneId, Boolean propagate) {
+    public static void updateGene(Integer entrezGeneId, Boolean propagate, Boolean skipGeneUpdateDelay) {
```
JiaoJiao123: Looks like skipGeneUpdateDelay is never used in any implementation.

zhx828: Ah yes, let me clean it up

PR #925, opened 2018-01-30
Update test; Fix treatment issues when treatment issues

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/TreatmentUtils.java
```diff
@@ -54,26 +54,24 @@ private TreatmentUtils() {
...
-        StringBuilder sb = new StringBuilder();
-        sorted = sorted == null ? false : sorted;
-
+    public static String getTreatmentName(Set<Treatment> treatments) {
         List<String> treatmentNames = new ArrayList<>();
-        for (Treatment treatment : treatments) {
+        List<Treatment> sortedTreatment = new ArrayList<>(treatments);
+        Collections.sort(sortedTreatment, new Comparator<Treatment>() {
+            public int compare(Treatment t1, Treatment t2) {
+                return t1.getPriority() - t2.getPriority();
```
JiaoJiao123: should we do the null check on the priority value?

PR #926, opened 2018-01-31
Fix issue related to not able to change treatmetnname 

zhx828:
Signed-off-by: Hongxin Zhang <hongxin@cbio.mskcc.org>


core/src/main/java/org/mskcc/cbio/oncokb/model/Evidence.java
```diff
@@ -371,13 +371,14 @@ public Integer getHighestTreatmentPriority() {
...
     }
 
-    public void setTreatments(Set<Treatment> treatments) {
+    public void setTreatments(List<Treatment> treatments) {
         if (treatments != null) {
-            for (Treatment treatment : treatments) {
-                treatment.setEvidence(this);
+            for (int i = 0; i < treatments.size(); i++) {
+                treatments.get(i).setDrugs(treatments.get(i).getDrugs());
+                treatments.get(i).setPriority(i);
```
JiaoJiao123: Should we use index + 1 to set the priority value?

PR #940, opened 2018-02-09
fix the geneStatus issue

JiaoJiao123:
The issue was caused by two things. 

One is the previous geneStauts reference is wrong. It needs to be specified based on level. 

The other is the treatment $index is wrong which caused the 4 angle icons for treatment section not displaying correctly. In this PR, we use indexOf to get the precise index instead of relying on the $index passing from gene.html

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -2521,13 +2521,13 @@ angular.module('oncokbApp')
                 var index = parent.indexOf(item);
                 switch (type) {
                 case 'top':
-                    if (index === 0) {
+                    if (index <= 1) {
                         return false;
                     } else {
                         return true;
                     }
                 case 'bottom':
-                    if (index === parent.length - 1) {
+                    if (index >= parent.length - 2) {
                         return false;
                     } else {
                         return true;
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
                 mutationEffect.addOn.setText('');
             };
 
-            $scope.move = function(driveList, index, moveIndex, event) {
+            $scope.displayAngle = function (parent, item, type) {
+                var index = parent.indexOf(item);
+                switch (type) {
+                case 'top':
+                    if (index === 0) {
+                        return false;
+                    } else {
+                        return true;
+                    }
+                case 'bottom':
+                    if (index === parent.length - 1) {
+                        return false;
+                    } else {
+                        return true;
+                    }
+                case 'up':
+                    if (index === 0) {
+                        return false;
+                    } else {
+                        return true;
+                    }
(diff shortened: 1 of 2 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
...
+                    } else {
+                        return true;
+                    }
+                case 'down':
+                    if (index === parent.length - 1) {
+                        return false;
+                    } else {
+                        return true;
+                    }
+                }
```
zhx828: Let's return a default `false` here.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
                 mutationEffect.addOn.setText('');
             };
 
-            $scope.move = function(driveList, index, moveIndex, event) {
+            $scope.displayAngle = function (parent, item, type) {
```
zhx828: Let's call this function `displayMoveIcon` even though the font-awesome calls them as angle icons

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
...
+                var index = parent.indexOf(item);
```
zhx828: Let's do a quick check on parent and item. Return false if the type is incorrect

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
...
+                switch (type) {
+                case 'top':
```
zhx828: The original logic is `index > 1`

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
...
+                    if (index === 0) {
+                        return false;
+                    } else {
+                        return true;
+                    }
+                case 'bottom':
+                    if (index === parent.length - 1) {
```
zhx828: The original logic is `index < parent.length - 2`

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2514,17 +2514,78 @@ angular.module('oncokbApp')
...
+                case 'down':
+                  moveIndex = index+2;
+                  break;
+                case 'top':
+                  moveIndex = 0;
+                  break;
+                case 'bottom':
+                  moveIndex = driveList.length;
+                  break;
+                }
```
zhx828: Technically, we should have a default as well. But optional from my opinion in this function.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: Current implementations are equivalent to original logic.

PR #945, opened 2018-02-13
Add endpoints to annotate genomic data based on passed in OncoKB variants

zhx828:
This solves issue https://github.com/oncokb/matchminer-curate/issues/21
https://github.com/oncokb/matchminer-curate/issues/44
https://github.com/oncokb/matchminer-curate/issues/45

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
@@ -1,8 +1,8 @@
 package org.mskcc.cbio.oncokb.api.pvt;
 
 import io.swagger.annotations.ApiParam;
-import org.mskcc.cbio.oncokb.apiModels.MatchVariantRequest;
 import org.mskcc.cbio.oncokb.apiModels.MatchVariant;
+import org.mskcc.cbio.oncokb.apiModels.MatchVariantRequest;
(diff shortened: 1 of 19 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
```diff
@@ -142,4 +148,73 @@
...
+            example = example.trim();
+            Alteration exampleVariant = new Alteration();
+            exampleVariant.setGene(gene);
+            exampleVariant.setAlteration(example);
+            exampleVariant.setName(example);
+            AlterationUtils.annotateAlteration(exampleVariant, example);
+
+            LinkedHashSet<Alteration> relevantAlterations = new LinkedHashSet<>();
+            if (isGeneralAlteration) {
+                relevantAlterations = alterationBo.findRelevantAlterations(exampleVariant, true);
```
JiaoJiao123: If the general mutation is not specifically curated, it won't be returned to the relevant alteration list, which will cause the mismatch. 
For example: utils/match/variant?hugoSymbol=EGFR&variant=Gain-of-function Mutations&examples=Amplification 
It will return {"Amplification": false}, which is wrong because Amplification is curated as Gain-of-function

zhx828:
@JiaoJiao123 thanks, they are all good points and have been addressed.

JiaoJiao123:
@zhx828 Looks good to me now

PR #946, opened 2018-02-14
clean up obsolete and red hand related code

JiaoJiao123:
The battery and red hand features have been removed from the curation platform for a while. It is time to clean up all the old features related code such as Obsolete/Unobsolete in Review Mode, Obsolete and Red hand options in gene data import, and Checking Variants by Status in tools page.


web/yo/app/scripts/services/stringutils.js
```diff
@@ -879,206 +879,174 @@ angular.module('oncokbApp')
...
-                    if (!(excludeObsolete && e.shortSummary_eStatus && e.shortSummary_eStatus.has('obsolete') && e.shortSummary_eStatus.get('obsolete') === 'true')) {
-                        _mutation = combineData(_mutation, e, ['shortSummary', 'oncogenic'], false, excludeComments, onlyReviewedContent);
-                    }
-                    if (!(excludeObsolete && e.oncogenic_eStatus && e.oncogenic_eStatus.has('obsolete') && e.oncogenic_eStatus.get('obsolete') === 'true')) {
-                        _mutation = combineData(_mutation, e, ['description', 'short', 'effect'], excludeObsolete, excludeComments, onlyReviewedContent);
-                        if (e.effect_uuid) {
-                            _mutation.effect_uuid = validUUID(e.effect_uuid);
-                        }
-                        _mutation.effect_review = getReview(e.effect_review);
+                if (onlyReviewedContent && e.name_review.get('added') == true || !onlyReviewedContent && e.name_review.get('removed') == true) return;
```
zhx828: `&& e.name_review.get('removed') == true` why we have new logic?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2397,7 +2239,7 @@ angular.module('oncokbApp')
 
             $scope.getAllCitations = function() {
                 var results = [];
-                var geneData = JSON.stringify(stringUtils.getGeneData(this.gene, true, true, true));
+                var geneData = JSON.stringify(stringUtils.getGeneData(this.gene, true, false));
```
zhx828: Why we change from true to false?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2464,7 +2306,7 @@ angular.module('oncokbApp')
             $scope.specifyAnnotation = function() {
                 var annotationLocation = {};
                 setAnnotationResult(annotationLocation, fetchResults(FindRegex.result(this.gene.background.text)), 'Gene Background');
-                var mutations = stringUtils.getGeneData(this.gene, false, false, false, false, true).mutations;
+                var mutations = stringUtils.getGeneData(this.gene, true, false).mutations;
```
zhx828: We only removed `excludeObsolete` and `excludeRedHands`, but why there are three parameters been removed here?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -2656,7 +2498,7 @@ angular.module('oncokbApp')
             };
 
             $scope.generatePDF = function() {
-                jspdf.create(stringUtils.getGeneData(this.gene, true, true, true));
+                jspdf.create(stringUtils.getGeneData(this.gene, true, false));
```
zhx828: Same here

web/yo/app/scripts/services/stringutils.js
JiaoJiao123: This logic is to apply onlyReviewedContent option to removed sections, which were somehow not included before. For instance, if a tumor is removed but not approved yet, we should still import this tumor if onlyReviewedContent is set to true. In the other way, that tumor will not be imported if onlyReviewedContent is set to false.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: Previously, getGeneData() function took 5 parameters, which means the last parameter onlyReviewedContent is using the default false value.

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: Previously, getGeneData() took 5 parameters and it was a mistake to pass 6

web/yo/app/scripts/controllers/gene.js
JiaoJiao123: same with above

PR #954, opened 2018-02-20
Sort treatment evidence in evidence endpoints

zhx828:
This solves issue #928

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/CustomizeComparator.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/CustomizeComparator.java
@@ -1,0 +1,108 @@
+package org.mskcc.cbio.oncokb.util;
+
+import org.mskcc.cbio.oncokb.model.Evidence;
+import org.mskcc.cbio.oncokb.model.IndicatorQueryTreatment;
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/CustomizeComparator.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+        });
+    }
+
+    private static int compareKitTreatmentOrder(String n1, String n2) {
+        int i1 = KIT_TREATMENT_ORDER.indexOf(n1);
+        int i2 = KIT_TREATMENT_ORDER.indexOf(n2);
+        if (i1 == i2)
+            return 0;
+        if (i1 == -1)
```
JiaoJiao123: When both of n1 and n2 are not in the KIT_TREATMENT_ORDER, should we return the result in alphabetical order?

zhx828: nope, return 0. It will keep the original order.

JiaoJiao123: ok I see

PR #955, opened 2018-02-21
Include mutation effect in indicator/search endpoint

zhx828:
This solves #913 #915

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
@@ -1,5 +1,6 @@
 package org.mskcc.cbio.oncokb.util;
 
+import org.mskcc.cbio.oncokb.apiModels.References;
 import org.mskcc.cbio.oncokb.model.*;
 import org.mskcc.cbio.oncokb.model.oncotree.TumorType;
 import org.w3c.dom.Document;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
```diff
@@ -141,6 +142,19 @@ public static MutationEffect findHighestMutationEffect(Set<MutationEffect> mutat
         return index == 100 ? null : PRIORITIZED_MUTATION_EFFECTS.get(index);
     }
 
+    public static IndicatorQueryMutationEffect findHighestMutationEffectByEvidence(Set<Evidence> evidences) {
+        Integer index = 100;
+        IndicatorQueryMutationEffect indicatorQueryMutationEffect = new IndicatorQueryMutationEffect();
+        for (Evidence evidence : evidences) {
+            MutationEffect mutationEffect = MutationEffect.getByName(evidence.getKnownEffect());
+            if (PRIORITIZED_MUTATION_EFFECTS.indexOf(mutationEffect) < index) {
```
JiaoJiao123: Shouldn't we update the "index" value to make the comparison valid? In this case, it is alway 100

zhx828: Ah yeah yeah, thanks!

PR #957, opened 2018-02-21
Improve performance based on JProfiler

zhx828:
1. Only parse string to integer when query by level
2. Try to avoid addAll when there is only one gene among alterations
3. Avoid converting set to list when Collection could be used as
parameter

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
@@ -271,10 +271,7 @@ public class EvidenceUtils {
         List<Evidence> evidences = new ArrayList<>();
 
         if (CacheUtils.isEnabled()) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
```diff
@@ -1000,4 +997,19 @@ public int compare(Evidence x, Evidence y) {
...
         return evidences;
     }
+
+    public static Set<Evidence> getAllEvidencesByAlterationsGenes(Collection<Alteration> alterations) {
+        Set<Gene> genes = new HashSet<>();
+        Set<Evidence> evidences = new HashSet<>();
+        for (Alteration alteration : alterations) {
+            genes.add(alteration.getGene());
+        }
+        if (genes.size() == 1) {
```
JiaoJiao123: I don't think you need to list this condition separately. It will be handled well in the for loop below.

PR #961, opened 2018-02-22
changes based on feedback for query reviewed data

JiaoJiao123:
It is about some feedback from Sarah. Mostly about adding 3 extra columns to gene type evidence.

### DIFF
--- a/web/yo/app/scripts/controllers/reportgenerator.js
+++ b/web/yo/app/scripts/controllers/reportgenerator.js
@@ -300,7 +300,7 @@ angular.module('oncokbApp')
                                 hugoSymbol: item.hugoSymbol
                             };
                         });
-                        var geneWithVariants = [];
+                        var geneWithVariants = {};
                         DatabaseConnector.lookupVariants(variantLookupBody).then(function(result) {
                             _.each(result, function(items) {
                                 var tempObj = {};
@@ -314,7 +314,7 @@ angular.module('oncokbApp')
                                             deletion: false,
                                             amplification: false
                                         };
-                                        geneWithVariants.push(item.gene.hugoSymbol);
+                                        geneWithVariants[item.gene.hugoSymbol] = true;
                                     }
                                     if (item.alteration === 'Truncating Mutations') {
                                         tempObj.truncatingMutations = true;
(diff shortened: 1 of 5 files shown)


web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -295,14 +295,55 @@ angular.module('oncokbApp')
 
                 DatabaseConnector.getReviewedData($scope.reviewedData[$scope.evidenceType].evidenceTypes).then(function(response) {
                     if ($scope.evidenceType === 'geneType') {
-                        _.each(response, function(item) {
-                            $scope.reviewedData.geneType.body.push({
-                                gene: item.hugoSymbol,
-                                oncogene: item.oncogene,
-                                tsg: item.tsg
+                        var variantLookupBody = _.map(response, function(item) {
```
zhx828:
```
_.map(_.uniq(_.pluck(response, 'hugoSymbol')), function(hugoSymbol){
    return {
        hugoSymbol: hugoSymbol
    }
})
```

web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -295,14 +295,55 @@ angular.module('oncokbApp')
...
+                                    if (item.alteration === 'Amplification') {
+                                        tempObj.amplification = true;
+                                    }
+                                });
+                                if (!_.isEmpty(tempObj)) {
+                                    $scope.reviewedData.geneType.body.push(tempObj);
+                                }
                             });
+                            _.each(response, function(item) {
+                                if (geneWithVariants.indexOf(item.hugoSymbol) === -1) {
```
zhx828: If this is the only place using `geneWithVariants`, suggest changing to object implementation for better performance. And this should be used for similar purpose across the project.

JiaoJiao123: Good point!

web/yo/app/scripts/controllers/reportgenerator.js
JiaoJiao123: _.map can take object directly and access as key-value pairs, it is unnecessary to form an array first

zhx828: The purpose is getting the unique hugoSymbol

JiaoJiao123: I thought there is no duplication in the genes list returned by the genes API. You think there might be?

zhx828: ... oh ok. You nested genes call in here as well. If that's the case, no. There is no dup.

Issue #918, opened 2018-01-26
Button: check all truncating mutations are in tumor suppressor genes


zhx828:
- [x] Endpoint to return the list of genes contains truncating mutations(containing certain variant)
- [x] Curation platform to add a button to trigger the API call, and check whether all genes are the tumor suppressor. Check the TSG property.

PR #962, opened 2018-02-23
Validate Truncating Mutations Close Issue #918

JiaoJiao123:


web/yo/app/scripts/controllers/reportgenerator.js
```diff
@@ -462,4 +462,58 @@ angular.module('oncokbApp')
...
+                DatabaseConnector.getReviewedData('geneType').then(function(response) {
+                    var geneTypes = {};
+                    var tempHugo = '';
+                    _.each(response, function(item) {
+                        geneTypes[item.hugoSymbol] = {
+                            oncogene: item.oncogene,
+                            tsg: item.tsg
+                        };
+                    });
+                    var variantCallBody = _.map(_.keys(geneTypes), function(hugoSymbol) {
```
zhx828: You can do this at line 480 since there is no dup.

PR #963, opened 2018-02-23
Get persistent drug object

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -387,6 +387,17 @@ private static void setAllAlterations() {
         return drugs;
     }
 
+    public static Drug getPersistentDrug(Drug drug) {
+        if (drug == null)
+            return null;
+
+        for (Drug persistent : drugs) {
```
JiaoJiao123: Is it because the performance of contains function is low that you switched to this manner?

zhx828: I need to get the persistent object. Contain only tells me whether the drug is exist

JiaoJiao123: I see

PR #969, opened 2018-03-01
Do not propagate alternative alleles Neutral/Likely Neutral mutation effect

zhx828:
This solves #870

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
@@ -157,6 +157,17 @@ public class MainUtils {
         return indicatorQueryMutationEffect;
     }
 
+    public static IndicatorQueryMutationEffect setToAlternativeAlleleMutationEffect(IndicatorQueryMutationEffect indicatorQueryMutationEffect) {
+        if (indicatorQueryMutationEffect != null && indicatorQueryMutationEffect.getMutationEffect() != null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
```diff
@@ -157,6 +157,17 @@ public static IndicatorQueryMutationEffect findHighestMutationEffectByEvidence(S
         return indicatorQueryMutationEffect;
     }
 
+    public static IndicatorQueryMutationEffect setToAlternativeAlleleMutationEffect(IndicatorQueryMutationEffect indicatorQueryMutationEffect) {
+        if (indicatorQueryMutationEffect != null && indicatorQueryMutationEffect.getMutationEffect() != null) {
+            MutationEffect mutationEffect = indicatorQueryMutationEffect.getMutationEffect();
+            MutationEffect likeME = MutationEffect.getByName("Likely " + mutationEffect.getMutationEffect().replaceAll("(?i)likely", "").trim());
+            if (likeME == null || likeME.equals(MutationEffect.LIKELY_NEUTRAL))
```
JiaoJiao123: If I understand correctly, likeME == null means the mutationEffect is inconclusive? If that is the case, could you please add some comment there?

zhx828: Sure

PR #971, opened 2018-03-02
 Fix issue evidence lookup does not return evidence when only hugoSymbol specified

zhx828:
This closes issue #938

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceTypeUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceTypeUtils.java
@@ -1,0 +1,66 @@
+package org.mskcc.cbio.oncokb.util;
+
+import org.mskcc.cbio.oncokb.model.EvidenceType;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceTypeUtils.java
```diff
@@ -0,0 +1,66 @@
...
+        evidenceTypes.add(EvidenceType.GENE_BACKGROUND);
+        return evidenceTypes;
+    }
+
+    public static Set<EvidenceType> getMutationEvidenceTypes() {
+        Set<EvidenceType> evidenceTypes = new HashSet<>();
+        evidenceTypes.add(EvidenceType.ONCOGENIC);
+        evidenceTypes.add(EvidenceType.MUTATION_EFFECT);
+        evidenceTypes.add(EvidenceType.VUS);
+        evidenceTypes.add(EvidenceType.MUTATION_SUMMARY);
```
JiaoJiao123: We don't use MUTATION_SUMMARY any more right? The mutation summary is contained in the MUTATION_EFFECT type evidence description.

zhx828: We haven't remove MUTATION_SUMMARY from the system yet. Will do in the same batch.

Issue #948, opened 2018-02-16
Queue Model


JiaoJiao123:
1: Transfer Queues data to Firestore
2: Modify current code base to connect with Firestore
3: Firestore queues data backup

JiaoJiao123:
REMINDER: **Remove** Queue model in the additional folder the moment release this change

PR #985, opened 2018-03-09
introduce firebase and integrate with current curation queue code base

JiaoJiao123:
Close Issue #948

### DIFF
--- a/web/yo/app/scripts/app.js
+++ b/web/yo/app/scripts/app.js
@@ -390,6 +390,7 @@ OncoKB.setUp = function(object) {
 OncoKB.keyMappings = {type: {TSG: '', OCG: ''}};
 
 OncoKB.initialize = function() {
+    firebase.initializeApp(OncoKB.config.firebaseConfig);
     var nonSetUp = ['TI'];
     var keys = window._.keys(OncoKB.curateInfo);
     var keysL = keys.length;
--- a/web/yo/app/scripts/app.js
+++ b/web/yo/app/scripts/app.js
@@ -503,7 +503,8 @@ var oncokbApp = angular.module('oncokbApp', [
     'datatables',
     'datatables.bootstrap',
     'localytics.directives',
-    'ui.sortable'
+    'ui.sortable',
+    'firebase'
 ])
     .value('user', {
         name: 'N/A',
--- a/web/yo/app/scripts/directives/curationQueue.js
+++ b/web/yo/app/scripts/directives/curationQueue.js
@@ -144,31 +144,6 @@ angular.module('oncokbApp')
                     }
                     return result;
                 };
-                $scope.generateQueuesList = function(hugoSymbol) {
-                    if (!$rootScope.queuesData.has(hugoSymbol)) {
-                        $rootScope.queuesData.set(hugoSymbol, $rootScope.queuesModel.createList());
-                    }
-                    _.each($rootScope.queuesData.get(hugoSymbol).asArray(), function(item) {
-                        $scope.queue.push({
-                            hugoSymbol: hugoSymbol,
-                            article: item.get('article'),
-                            pmid: item.get('pmid'),
-                            pmidString: 'PMID: ' + item.get('pmid'),
-                            link: item.get('link'),
-                            variant: item.get('variant'),
-                            mainType: item.get('mainType'),
-                            subType: item.get('subType'),
-                            section: item.get('section'),
-                            addedBy: item.get('addedBy'),
-                            addedAt: item.get('addedAt'),
-                            curated: item.get('curated'),
-                            curator: item.get('curator'),
-                            comment: item.get('comment'),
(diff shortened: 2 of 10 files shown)


web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -351,41 +331,80 @@ angular.module('oncokbApp')
...
-                    queueItem.comment = queueModelItem.get('comment');
-                    if ($scope.data.resendEmail) {
-                        $scope.sendEmail(queueItem, queueModelItem);
+                    var hugoSymbol = queueItem.hugoSymbol;
+                    var currentQueues = $scope.getQueuesByGene(hugoSymbol);
+                    var item = {};
+                    for (var i = 0; i < currentQueues.length; i++) {
+                        if (currentQueues[i].addedAt === queueItem.addedAt) {
+                            item = angular.copy(queueItem);
+                            item['link'] = $scope.input.link;
```
zhx828: FYI, you could use item.link to simplify(also recommended).

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -584,6 +596,21 @@ angular.module('oncokbApp')
...
+                    return $rootScope.firebaseQueues[hugoSymbol] ? angular.copy($rootScope.firebaseQueues[hugoSymbol].queue) : [];
+                };
+                $scope.updateQueueInDB = function (hugoSymbol, updatedQueues) {
+                    var defer = $q.defer();
+                    firebase.database().ref('Queues/' + hugoSymbol).set({
+                        queue: updatedQueues
+                    }).then(function(result) {
+                        defer.resolve('success');
+                    }).catch(function(error) {
+                        dialogs.error('Error', 'Fail to save changes to database. Please contact developer!');
```
zhx828: This error message should be saved somewhere in the global vars, then it can be used everywhere.

web/yo/app/scripts/app.js
```diff
@@ -390,6 +390,7 @@ OncoKB.setUp = function(object) {
 OncoKB.keyMappings = {type: {TSG: '', OCG: ''}};
 
 OncoKB.initialize = function() {
+    firebase.initializeApp(OncoKB.config.firebaseConfig);
```
zhx828: Where this firebase var imported? From window object? Could you register it when initializing the app? Similarly to underscore.

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -212,46 +212,32 @@ angular.module('oncokbApp')
...
-                    }
-                    $rootScope.queuesData.get(hugoSymbol).push(item);
-                    var queueItem = {
-                        article: item.get('article'),
-                        pmid: item.get('pmid'),
-                        pmidString: 'PMID: ' + item.get('pmid'),
-                        link: item.get('link'),
-                        variant: item.get('variant'),
-                        mainType: item.get('mainType'),
-                        subType: item.get('subType'),
```
zhx828: why are those logics removed?

web/yo/app/scripts/directives/curationQueue.js
```diff
@@ -212,46 +212,32 @@ angular.module('oncokbApp')
...
-                        notified: item.get('notified'),
-                        hugoSymbol: hugoSymbol
-                    };
-                    $scope.queue.push(queueItem);
-                    if (item.get('curator')) {
-                        $scope.sendEmail(queueItem);
+                        item['article'] = $scope.input.article;
                     }
+                    currentQueues.push(item);
+                    $scope.updateQueueInDB(hugoSymbol, currentQueues).then(function(result) {
```
zhx828: This is what I was talking about that everytime you change the queue content, you need to manually call the updateQueueInDB method. Isn't there a way to keep the current functionality?

web/yo/app/scripts/directives/curationQueue.js
JiaoJiao123: I see your point to reuse redundant message variable. But right now, this message is unique. I will put it in a global varible when necessary.

web/yo/app/scripts/directives/curationQueue.js
JiaoJiao123: Previously we need two variable item and queueItem, one is JSON object, the other is the Map object for the queue model in Google Drive. Now we don't need the second one any more so the logic is removed.

web/yo/app/scripts/factories/VariantFactory.js
```diff
@@ -86,9 +86,14 @@ angular.module('oncokbApp').factory('DriveOncokbInfo', ['$http', 'OncoKB', funct
         return $http.get('data/oncokbInfo.json');
     }
 
+    function getFirebasePermission() {
+        return $http.get('data/firebaseUserPermission.json');
```
zhx828: I thought we are loading user permission from firebase?

Issue #999, opened 2018-03-14
Include tissue level when checking the tumor form


zhx828:
Currently, we use list main types to figure out whether it is a solid tumor or liquid tumor. But it may change frequently.

PR #1002, opened 2018-03-14
Use tissue + main type to determine whether it is liquid tumor

zhx828:
All nodes under tissues "Lymph", "Blood", "Lymphoid", "Myeloid" will be
treated as liquid tumor
This solves issue #999 and #1000


core/src/main/java/org/mskcc/cbio/oncokb/util/TumorTypeUtils.java
```diff
@@ -31,14 +31,19 @@
...
     }};
 
-    private static List<TumorType> allOncoTreeSubtypes =
-        getOncoTreeSubtypesByCancerTypesFromSource(allOncoTreeCancerTypes);
+    private static List<TumorType> allOncoTreeSubtypes = getOncoTreeSubtypesFromSource();
     private static Map<String, TumorType> allNestedOncoTreeSubtypes = getAllNestedOncoTreeSubtypesFromSource();
     private static Map<String, List<TumorType>> questTumorTypeMap = null;
     private static Map<String, List<TumorType>> cbioTumorTypeMap = null;
+    private static final ImmutableList<String> LiquidTumorTissues = ImmutableList.of(
+        "Lymph", "Blood", "Lymphoid", "Myeloid"
```
JiaoJiao123: So the tissue name changed from BLOOD to Blood?

zhx828: The name is always Blood. BLOOD is the code, but we don't use it on the tissue level.

PR #1019, opened 2018-03-23
Load History data from Firebase

victoria34:
Fixed issue: [#951]

### DIFF
--- a/web/yo/app/scripts/services/databaseconnector.js
+++ b/web/yo/app/scripts/services/databaseconnector.js
@@ -22,6 +22,7 @@ angular.module('oncokbApp')
         'ApiUtils',
         'PrivateApiUtils',
         'user',
+        'storage',
         function($timeout,
                  $q,
                  $rootScope,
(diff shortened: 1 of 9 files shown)


JiaoJiao123:
screen shot 2018-03-26 at 3 01 21 pm

JiaoJiao123:
@victoria34 Firstly, the Gene column is empty. 
Secondly, remove the history model back up since it is not needed any more

web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -719,35 +721,35 @@ angular.module('oncokbApp')
...
-                        function(error) {
-                            console.log('fail to send old history to oncokb dev account', error);
-                        }
-                    );
-                }
-                apiHistory.push({
-                    admin: user.name,
-                    timeStamp: new Date().getTime(),
-                    records: historyData
+            function updateHistory(hugoSymbol, historyData) {
```
JiaoJiao123: Passing hugoSymbol down to here is not a good way to share data. Please assign current gene hugoSymbol to a rootScope variable, just like the way how $rootScope.model is assigned

JiaoJiao123:
History related factor has already finished. This PR is not relevant anymore.

Issue #1036, opened 2018-04-06
Support HEAD method


zhx828:
https://sentry.io/memorial-sloan-kettering/oncokb/issues/517589939/?query=is:unresolved

PR #1047, opened 2018-05-03
Approve sentry to exclude certain errors; Support HEAD method to all GET methods

zhx828:
This solves issue #1036


web/src/main/java/org/mskcc/cbio/oncokb/config/MvcConfiguration.java
```diff
@@ -72,6 +77,28 @@ protected ApiInfo apiInfo() {
...
+                                                 HttpServletResponse response,
+                                                 Object handler,
+                                                 Exception ex) {
+                Throwable rootCause = ex;
+
+                while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {
+                    rootCause = ex.getCause();
+                }
+
+                if (!rootCause.getMessage().contains("Broken pipe")
```
JiaoJiao123: Should we make the contains check case insensitive?

PR #1077, opened 2018-06-08
Fix gene type old content not checked issue

victoria34:
Issue: #1080


JiaoJiao123:
The fix is not designed properly. Already explained to Jing, and the fix is done in another PR

PR #1161, opened 2018-07-05
Think gene type without key_reviewed as checked by default.

victoria34:


JiaoJiao123:
Updated code is integrated into other PR

PR #1265, opened 2018-07-23
Fix same tumor type issue

victoria34:
Issue #1246


web/yo/app/scripts/services/mainutils.js
```diff
@@ -87,15 +87,42 @@ angular.module('oncokbApp')
...
                 }
             });
             list = _.uniq(list).sort();
             return list.join(', ');
         }
+
+        function hasDuplicateCancerTypes(cancerTypes) {
+            var result = false;
+            if (!cancerTypes) {
+                result = false;
```
JiaoJiao123: change to "return false"
otherwise it will keep going to the _.some() function and throw error

Issue #1256, opened 2018-07-18
Cursor keeps going back to the beginning of the paragraph when I'm trying to type!


debyani-chakravarty:
Please give as much information as possible before submitting your issue - Thank you

Issue: 

Example:

Screenshot:

PR #1275, opened 2018-07-24
Remove default from updateOn

victoria34:
Blur means ng-model will be updated when blur event happened. Default means ng-model will be updated in realtime.

Issue #1256


zhx828:
@victoria34 but you do want to update the model in realtime though, so that's why we set a delay. Blur only works when the box lost the focus. @JiaoJiao123 you probably have more experience here.

PR #1278, opened 2018-07-24
Sort relevant alterations

zhx828:
1. Sort based on the length of the overlap if the relevant is range mutation, the bigger overlap range will be prioritized.
2. Sort based on the length of the range mutation, the shorter range mutation will be prioritized.

Checked the system that we do support range splice mutation
Related to #1257

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -690,6 +690,27 @@ public final class AlterationUtils {
         });
     }
 
+    public static void sortAlterationsByTheRange(List<Alteration> alterations) {
+        Collections.sort(alterations, new Comparator<Alteration>() {
+            @Override
+            public int compare(Alteration a1, Alteration a2) {
+                if (a1.getProteinStart() == null || a1.getProteinEnd() == null) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -690,6 +690,34 @@ public int compare(Alteration a1, Alteration a2) {
...
+                }
+
+                int overlap1 = Math.min(a1.getProteinEnd(), proteinEnd) - Math.max(a1.getProteinStart(), proteinStart);
+                int overlap2 = Math.min(a2.getProteinEnd(), proteinEnd) - Math.max(a2.getProteinStart(), proteinStart);
+
+                if (overlap1 == overlap2) {
+                    int diff = a1.getProteinEnd() - a1.getProteinStart() - (a2.getProteinEnd() - a2.getProteinStart());
+                    return diff == 0 ? a1.getAlteration().compareTo(a2.getAlteration()) : diff;
+                } else {
+                    return overlap2 - overlap1;
```
victoria34: How do you ensure that overlap2 > overlap1?

zhx828: I don't need to ensure 2 > 1. If it is bigger then it will return a positive number which 2 will have a higher priority. vice versa

PR #1287, opened 2018-07-25
solve cursor moving issue

JiaoJiao123:
Two things to highlight:
a) 2 seconds was changed to 0.2 seconds to reduce the risk of a curator losing data in the case that, multiple people edit on the same section at the same time
b) The 2 seconds debouce delay we added should already to keep the cursor stable. But since we re-assign the data[key] value after removing CSS style, the cursor moves to the very beginning. The solution is to only run the function to strip CSS when copy and paste happens.


JiaoJiao123:
Solves Issue #1256

zhx828:
@JiaoJiao123 in the previous implementation, even I didn't paste anything, it will still causing the cursor jumped to the beginning. I decreased the time to 1s before, it didn't work. Just curious how this pr would solve the issue.

JiaoJiao123:
@zhx828 In the previous implementation, removing CSS function is called regardless of typing or pasting. And that function is the cause for the jumping cursor after we implemented the 2s debounce delay. The solution here is to only call the function to remove CSS when the changes are from pasting.

Issue #1260, opened 2018-07-18
The actionable alterations download file should be sorted


zhx828:
The order should be Level, Gene, Cancer Type, Alteration

PR #1299, opened 2018-07-31
Sort All Variants, Actionable Variants file

zhx828:
Actionable Variants: Level -> Gene -> Tumor Type -> Alteration
All Variants: Gene -> Alteration -> Oncogenicity -> Mutation Effect

This solves issue #1260

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
@@ -13,8 +13,6 @@ import org.springframework.stereotype.Controller;
 
 import java.util.*;
 
-;
-
 /**
  * Created by Hongxin on 10/28/16.
  */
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -3,15 +3,18 @@
...
 import org.mskcc.cbio.oncokb.util.GeneUtils;
 import org.mskcc.cbio.oncokb.util.MainUtils;
-import org.mskcc.cbio.oncokb.model.oncotree.TumorType;;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 
 import java.util.*;
+
+;
```
victoria34: What's this `;`?

zhx828: Oh, probably the autoformatting moved the top double semicolon down. I will update.

Issue #1121, opened 2018-06-28
Lower level treatment should not be shown in evidence call


zhx828:
<img width="455" alt="screen shot 2018-06-28 at 1 36 38 pm" src="https://user-images.githubusercontent.com/5400599/42053133-b28d0cc0-7ade-11e8-8c35-576ea6ba94b1.png">

The temsirolimus on level 4 shouldn’t be shown here

zhx828:
Portal should keep the same rule of the MSK reports,
Always show the highest level treatment if the level is one of 1, 2A, 3A, 4
If the highest is 2B or 3B and it has second line of therapy, should that as additional therapy

zhx828:
If the same treatment(A) appears in the string A, B then the whole drug family A, B will be ignored. But if A is in combo A+B, the evidence will be shown.

PR #1305, opened 2018-08-02
Filter out evidence with lower level

zhx828:
This solves issue #1121

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
@@ -62,7 +62,7 @@ public class EvidenceUtils {
                 (source != null ? ("&" + source) : "") +
                 "&" + evidenceTypes.toString() +
                 (levelOfEvidences == null ? "" : ("&" + levelOfEvidences.toString()));
-            if(matchedAlt == null) {
+            if (matchedAlt == null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
```diff
@@ -613,24 +613,57 @@ public static Oncogenicity getOncogenicityFromEvidence(Set<Evidence> evidences)
...
 
         for (Evidence evidence : evidences) {
             if (evidence.getTreatments() != null && evidence.getTreatments().size() > 0) {
-                String treatmentsName = TreatmentUtils.getTreatmentName(new HashSet<>(evidence.getTreatments()));
-                if (!maps.containsKey(treatmentsName)) {
-                    maps.put(treatmentsName, new HashSet<Evidence>());
+                List<String> treatments = TreatmentUtils.getTreatments(new HashSet<>(evidence.getTreatments()));
+
+                Map<String, Object> map = getBiggerItem(treatments, maps.keySet());
+                if ((boolean) map.get("exist")) {
```
victoria34: Just curious. Why do you convert it forcefully to boolean in here? I thought its value would only be True or False.

zhx828: The map has Object content, you need to force converting Object to boolean in order to pass the validation process.

Issue #1296, opened 2018-07-27
Review actions not save in History tables.


victoria34:
Please give as much information as possible before submitting your issue - Thank you

Issue: 
For gene KIT, there is a mutation with "Other Tumor Types".  It only has tumor type summary. When curator delete it and approved in review mode, the change doesn't reflect to MySQL and FIrebase History.

Example:

Screenshot:

PR #1306, opened 2018-08-03
Add new and old content for History in Firebase.

victoria34:
It may fix issue #1296.


web/yo/app/views/realtimeString.html
```diff
@@ -51,7 +51,7 @@
                             chosen allow-single-deselect="true"
                             ng-model="data[key]"
                             ng-options="opt.value as opt.label for opt in o">
-                        <option value="NA"></option>
```
zhx828: Explanation here?

victoria34: This is a bug fix since I run into issue https://github.com/harvesthq/chosen/issues/1942.

PR #1337, opened 2018-08-17
 Fix the tests based on the latest data

zhx828:


web/yo/app/scripts/services/databaseconnector.js
```diff
@@ -254,7 +255,7 @@ angular.module('oncokbApp')
             }
 
             function sendEmail(params, success, fail) {
-                if (testing) {
+                if (testing || inProduction) {
```
victoria34: It won't allow sending email in production right?

PR #1367, opened 2018-09-05
Fix url localStorage issue

### LINKED ISSUE #1364: The URL should overwrite the localstorage
When a user specifies the gene in the URL, it should overwrite the content saved in the localstorage. If the gene in the URL is a valid gene, then go to that gene page, otherwise, go to genes page. If there is no gene specified in the URL, read from localstorage.

victoria34:
Fix #1364

### DIFF
--- a/web/yo/app/scripts/app.js
+++ b/web/yo/app/scripts/app.js
@@ -181,19 +181,20 @@ angular.module('oncokbApp').run(
             $rootScope.$on('$routeChangeStart', function(event, next) {
                 var fromIndex = window.location.href.indexOf('/gene/');
                 var hugoSymbol = '';
+                var regex = /\/([^\/]+)\/?$/;
                 if (fromIndex !== -1) {
                     //When the curator left the gene page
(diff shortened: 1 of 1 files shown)


web/yo/app/scripts/app.js
```diff
@@ -183,20 +183,19 @@ angular.module('oncokbApp').run(
...
                     //When the curator left the gene page
-                    hugoSymbol = window.location.href.substring(fromIndex+6);
+                    hugoSymbol = window.location.href.match(/\/([^\/]+)\/?$/)[1];
                     window.localStorage.geneName = hugoSymbol;
                 }
                 var toIndex = $location.path().indexOf('/gene/');
                 if (toIndex !== -1) {
                     //When the curator enter the gene page
-                    hugoSymbol = $location.path().substring(toIndex+6);
+                    hugoSymbol = $location.path().match(/\/([^\/]+)\/?$/)[1];
```
zhx828: Let's make this regular expression a constant, it can be used in these three places. Or even add a util to parse the URL.

PR #1386, opened 2018-10-05
Update drug model and add DrugGroup

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/bo/DrugGroupBo.java
```diff
@@ -0,0 +1,20 @@
...
+     * @param groupName
+     * @return
+     */
+    DrugGroup findGroupByName(String groupName);
+
+    /**
+     * @param drug
+     * @return
+     */
+    List<DrugGroup> findGroupsByDrug(Drug drug);
```
jjgao: Do we also need `findDrugsInGroup` in order to match trials?

zhx828: I don't think so. The drug table should cover all drugs. Unless they may mention anything like `MEK inhibitors`?

zhx828: Right, we need that as well. For additional endpoints, I can add them when I work on the frontend features.

PR #1407, opened 2018-10-31
Switch to spotify/dockerfile-maven + webapp-runner

inodb:
Using webapp-runner provides an easier command line interface to tomcat imo.
Change the docker image also to mount at root instead of at /oncokb/ path.
Changed the docker generation to use spotify/dockerfile-maven, this is
recommended new way to generate docker files using maven.


inodb:
@ersinciftci 	☝️

ersinciftci:
Yeah I think it's better to have a Dockerfile that everyone can understand instead of having the same info in pom.xml

ersinciftci:
I thought it would make things easier if we didn't have a Dockerfile when I first made this change, but I don't think like that anymore :)

inodb:
@ersinciftci agreed - being able to see the Dockerfile is useful. I guess that's why spotify/docker-maven-plugin now recommends to use spotify/dockerfile-maven

PR #1414, opened 2018-11-14
Use Collections.unmodifiedSet instead of creating HashSet

zhx828:


zhx828:
This is a part of fixes for https://github.com/cBioPortal/cbioportal/issues/5167

PR #1420, opened 2018-11-21
Add Therapy

Yizhuzhu:
ModifyTumorTypes.html uses inline CSS. To keep consistent, I didn't change main.scss and use inline css too. Thank you for reviewing.

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -3223,19 +3223,15 @@ angular.module('oncokbApp')
             //The previous drugs?
             _.each(therapyuuid,function(element){
                 _.each(element,function(drug){
-                    // _.each(data.cancerTypes, function(cancerType) {
-                    //     if (cancerType.mainType) {
-                    //         firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationName +'/').once('value', function(snapshot){
-                    //             if (snapshot.hasChild(cancerType.mainType) == false){
-                    //                 firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationName + '/' + cancerType.mainType).set('');
-                    //             }
-                    //         });
-                    //     }
-                    //     if (cancerType.subtype) {
-                    //         firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationName +'/' + cancerType.mainType + '/' + cancerType.subtype).set('');
-                    //     }
-                    // });
-                    firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationUuid).set('');
+                    _.each(data.cancerTypes, function(cancerType) {
+                        if(cancerType.code!="")
+                            firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationUuid + '/' + cancerType.code + '/' + treatment.name_uuid).set('');
+                        else{
+                            firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationUuid + '/' + cancerType.mainType + '/' + treatment.name_uuid).set('');
+                        }
+                    });
+
+                    //firebase.database().ref('Map/' + drug + '/' + data.geneName + '/' + data.mutationUuid).set('');
                 });
             });
             // if (!data.tiRef.treatments) {
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -2988,12 +2988,13 @@ angular.module('oncokbApp')
         }]
     )
 
-    .controller('ModifyTherapyCtrl', function ($scope, $modalInstance, data, _, OncoKB, $rootScope, mainUtils, FirebaseModel, $timeout, DatabaseConnector){
+    .controller('ModifyTherapyCtrl', function ($scope, $modalInstance, data, _, OncoKB, $rootScope, mainUtils, FirebaseModel, $timeout, DatabaseConnector, loadFiles, $firebaseObject, $firebaseArray, $q){
         $scope.therapy = [[]];
         $scope.addTherapyError = false;
         // $scope.loadDrugs = function($query){
         //     return DatabaseConnector.getAllDrugs()
         //         .then(function(result){
+        //             console.log(result);
         //             $scope.addTherapyError = false;
         //             var drugs = result;
         //             return drugs.filter(function (drug) {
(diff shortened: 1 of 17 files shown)


web/yo/app/views/modifyTherapy.html
```diff
@@ -0,0 +1,38 @@
...
+            <div ng-click="addTherapy($index)" class="col-lg-10 col-md-10 col-sm-10 col-xm-10">
+                <tags-input ng-model="element.tags" display-property="drugName" replace-spaces-with-dashes="false" placeholder="  " add-from-autocomplete-only="true" on-tag-added="addDruginTherapy($tag, $index)" on-tag-removed="removeDruginTherapy($tag, $index)">
+                    <auto-complete source="loadDrugs($query)"
+                                   min-length="1"
+                                   load-on-focus="true"
+                                   load-on-empty="false"
+                                   template="drugsDropdown"></auto-complete>
+                </tags-input>
+            </div>
+            <button class="col-lg-1 col-md-1 col-sm-1 col-xm-1" ng-click="deleteTherapy($index)" style="background-color: white; border-color: white;"><i class="fa fa-remove" style="font-size:18px; margin-left:-10px; margin-top: 10px;"></i></button>
```
zhx828: Let's keep the style consistency, use the class='btn btn-default [maybe btn-sm?]'

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,115 @@ angular.module('oncokbApp')
...
+
+        $scope.removeDruginTherapy = function(tag,index){
+            var id = $scope.therapy[index].indexOf(tag.drugName);
+            if (id > -1) {
+                $scope.therapy[index].splice(id, 1);
+            }
+            addTherValiEle(index);
+        }
+
+        function addTherValiEle(index) {
```
zhx828: Let's use the full name for the function, it would be easier to understand

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,115 @@ angular.module('oncokbApp')
...
+        $scope.deleteTherapy = function(index){
+            if (index > 0){
+                $scope.therapy.splice(index, 1);
+            }
+        }
+
+        function isValidTreatment(indices, newTreatmentName) {
+            var isValid = true;
+            _.some(data.treatmentRef.treatments, function (treatment) {
+                if (treatment.name.toLowerCase() === newTreatmentName.toLowerCase()) {
```
zhx828: why toLowerCase?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,115 @@ angular.module('oncokbApp')
...
+
+        $scope.deleteTherapy = function(index){
+            if (index > 0){
+                $scope.therapy.splice(index, 1);
+            }
+        }
+
+        function isValidTreatment(indices, newTreatmentName) {
+            var isValid = true;
+            _.some(data.treatmentRef.treatments, function (treatment) {
```
zhx828: You could use _.find

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,115 @@ angular.module('oncokbApp')
...
+                }
+            });
+            if (!isValid) {
+                $scope.sameTherapy = true;
+            }
+            return isValid;
+        }
+
+        $scope.save = function(){
+            while($scope.therapy[$scope.therapy.length-1]==''){
```
zhx828: I don't suggest using while+splice on an array. Why not just combined with the each below?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,115 @@ angular.module('oncokbApp')
...
+                $scope.therapy.splice($scope.therapy.length-1, 1);
+            }
+            var therapyString = [];
+            var indices = data.indices;
+            if ($scope.sameElement === false){
+                _.each($scope.therapy,function(element){
+                    element = element.join(' + ').trim();
+                    therapyString.push(element);
+                });
+                var newTreatmentName = therapyString.join(', ').trim();
```
zhx828: Why trim again?

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,115 @@ angular.module('oncokbApp')
...
+                var treatment = new FirebaseModel.Treatment(newTreatmentName);
```
zhx828: @victoria34 could you pay attention to the below logic? Not familiar with that.

victoria34: @zhx828  The treatment is defined by Jiaojiao and it works well. Yi doesn't change this part of code.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,119 @@ angular.module('oncokbApp')
...
+        function valiElement(index) {
+            $scope.addTherapyError = false;
+            var therapyVali = [];
+            _.each($scope.therapy, function(element) {
+                element = element.sort();
+                element = element.join(' ').trim();
+                if (element!=='')
+                    therapyVali.push(element);
+            });
+            for(var i=0;i<therapyVali.length;i++){
```
victoria34: For line 3037 ~ 3043, I think you can simply this part of code. If you want to check if 2 arrays are exactly same, you can sort them first and then compare them directly.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3009,6 +2987,119 @@ angular.module('oncokbApp')
...
+            });
+            if (!isValid) {
+                $scope.therapyErrorMessage = "Same therapy exists.";
+                $scope.addTherapyError = true;
+            }
+            return isValid;
+        }
+
+        $scope.save = function(){
+            for(;$scope.therapy[$scope.therapy.length-1]=='';){
```
victoria34: What's this for loop used for?

victoria34: What's this for loop used for?

web/yo/app/scripts/controllers/gene.js
Yizhuzhu: It's not for checking if 2 arrays are exactly same. It's for checking if they contain same elements.

web/yo/app/scripts/controllers/gene.js
Yizhuzhu: To make sure that there is no blank therapy. For example, the user clicks the second element, the third element will show. If she didn't input anything in the second and third elements and didn't delete them, this function would help delete them.

PR #1429, opened 2018-12-05
Allow admin to create genes

victoria34:
screen shot 2018-12-05 at 5 24 33 pm


victoria34:
@zhx828 From line 193 ~ 219, I created multiple warning pop-ups for different error case, such as duplicate gene, fail to create a Meta for a gene and fail to create a gene.

PR #1450, opened 2019-01-16
Remove edit column and increase debounce time for cursor jumping issue

victoria34:


web/yo/app/views/realtimeString.html
```diff
@@ -8,7 +8,7 @@
                         ng-class="getInputClass()"
                         ng-model="data[key]"
                         ng-paste="trimCSS()"
-                        ng-model-options="{ updateOn: 'default blur', debounce: { 'default': 200, 'blur': 0 } }">
```
zhx828: We no longer need to specify `updateOn`?

victoria34: I think that's not necessary. When someone is editing content, it will block another user from editing for 2s. In this way, I think we don't need to update ng-model immediately when `blur` event is triggered. If we don't need to specify debounce time for `default` and `blur` events separately, `updateOn` can be removed.

Issue #1453, opened 2019-01-17
evidences endpoint does not give any information when gene is not specified


zhx828:
When using `https://oncokb.org/api/v1/evidences/lookup?source=oncotree&evidenceTypes=GENE_SUMMARY`, it returns empty array.

Issue #1452, opened 2019-01-16
Sort 3A over 2B in the treatment list also in portal


zhx828:

jjgao:
Should this apply to all cases?

PR #1457, opened 2019-01-23
List 3A over 2B when sorting treament list

zhx828:
This solves https://github.com/oncokb/oncokb/issues/1452 https://github.com/oncokb/oncokb/issues/1453

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
@@ -109,8 +109,16 @@ public class EvidenceUtils {
     }
 
     public static Set<Evidence> getEvidenceByEvidenceTypesAndLevels(Set<EvidenceType> types, Set<LevelOfEvidence> levels) {
-        Set<Alteration> alterations = AlterationUtils.getAllAlterations();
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceUtils.java
```diff
@@ -109,8 +109,16 @@
     }
 
     public static Set<Evidence> getEvidenceByEvidenceTypesAndLevels(Set<EvidenceType> types, Set<LevelOfEvidence> levels) {
-        Set<Alteration> alterations = AlterationUtils.getAllAlterations();
-        List<Evidence> evidences = EvidenceUtils.getEvidence(new ArrayList<>(alterations), types, levels);
+        List<Evidence> evidences = new ArrayList<>();
+        for (Evidence evidence : CacheUtils.getAllEvidences()) {
+            if (types != null && types.size() > 0 && !types.contains(evidence.getEvidenceType())) {
```
victoria34: Can you merge these 2 if statements? Or why don't you write in this way?
if (condition) {
   evidences.add(evidence);
} else {
    continue;
}

zhx828: I think it's the same. Separating the conditions makes the code more readable.

PR #1472, opened 2019-02-01
Map

Yizhuzhu:


web/yo/app/scripts/controllers/drugs.js
```diff
@@ -10,19 +10,49 @@ angular.module('oncokbApp')
                 }, function (error) {
                     deferred1.reject(error);
                 });
+                var deferred2 = $q.defer();
+                $scope.drugMap = {};
+                firebase.database().ref('Map').once('value').then(function (snapshot) {
```
victoria34: Why you use `once` instead of  `on`?

web/yo/app/scripts/controllers/drugs.js
```diff
@@ -31,24 +61,40 @@ angular.module('oncokbApp')
...
+                    modalError("Sorry", "Same name exists.", true, false, drug.uuid);
                 } else {
                     if (!newDrugName)
                         newDrugName = drug.drugName;
                     firebaseConnector.set('Drugs/' + drug.uuid + '/drugName', newDrugName);
                 }
             };
+
+            $scope.removeDrug = function (drug) {
+                firebaseConnector.ref('Map').once('value', function (snapshot) {
```
victoria34: In here, you don't need to refetch data from firebase. You can check if the uuid exists in _.keys($scope.drugMap) directly when you use `on` in previous comment, .

web/yo/app/scripts/controllers/drugs.js
```diff
@@ -31,24 +61,40 @@ angular.module('oncokbApp')
...
+            };
         }]
     )
-    .controller('ModalErrorCtrl', function ($scope, $modalInstance, data) {
+    .controller('ModalErrorCtrl', function ($scope, $modalInstance, data, firebaseConnector) {
         $scope.errorTitle = data.errorTitle;
         $scope.errorMessage = data.errorMessage;
+        $scope.sameName = data.sameName;
         $scope.deleteDrug = data.deleteDrug;
+        if(data.genes !== undefined){
```
victoria34: if (data.genes) // When data.genes=undefined, it will return false.

web/yo/app/scripts/controllers/drugs.js
```diff
@@ -31,24 +61,40 @@ angular.module('oncokbApp')
...
+            $scope.drugGenes = data.genes.join(', ');
+        }
         $scope.cancel = function () {
             $modalInstance.dismiss('canceled');
         };
         $scope.confirm = function () {
-            firebaseConnector.set('Drugs/' + data.drug.uuid, null);
+            firebaseConnector.set('Drugs/' + data.drugUuid, null);
```
victoria34: If you send `null` to firebase, firebase won't create anything.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1852,6 +1910,42 @@ angular.module('oncokbApp')
             $scope.getLastReviewedCancerTypesName = function (tumor) {
                 return mainUtils.getCancerTypesName(tumor.cancerTypes_review.lastReviewed);
             };
+
```
victoria34: removeTherapyInMap and saveToMap functions are so similar. Could you extract common part of them or merge them together? For example,
```
function example(type, ...) { 
                var indices = getIndexByPath(mapPath);
                var geneName = $scope.gene.name;
                var mutationName = $scope.gene.mutations[indices[0]].name;
                var cancerTypes = $scope.gene.mutations[indices[0]].tumors[indices[1]].cancerTypes;

                if (type === 'remove') {
                        removeTherapyInMap(keys, therapyUuid, mapPath);
                } else {  
                        saveToMap(therapyUuid, name_uuid, name, mapPath);
                } 
}
```

web/yo/app/scripts/controllers/drugs.js
Yizhuzhu: Yes, or I can use FirebaseConnector.remove here

web/yo/app/scripts/controllers/drugs.js
```diff
@@ -31,24 +61,40 @@ angular.module('oncokbApp')
 
             $scope.saveDrugName = function (newDrugName, drug) {
                 if (hasSameName(newDrugName, drug.uuid)) {
-                    modalError("Sorry", "Same name exists.", false, drug);
+                    modalError("Sorry", "Same name exists.", true, false, drug.uuid);
```
victoria34: modalError() has 6 parameters. In here, you only pass 5 so the last one is `null` by default?

web/yo/app/scripts/controllers/drugs.js
```diff
@@ -31,24 +61,40 @@ angular.module('oncokbApp')
...
         $scope.deleteDrug = data.deleteDrug;
+        if(data.genes){
+            $scope.drugGenes = data.genes.join(', ');
+        }
         $scope.cancel = function () {
             $modalInstance.dismiss('canceled');
         };
         $scope.confirm = function () {
-            firebaseConnector.set('Drugs/' + data.drug.uuid, null);
+            firebaseConnector.remove('Drugs/' + data.drugUuid);
```
victoria34: I think you should create a function called `removeDrug(uuid)` in firebaseconnector.js. You should only pass drug uuid to that function. In this way, we can prevent people write wrong path and mess up database.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1852,6 +1910,39 @@ angular.module('oncokbApp')
...
+            function changeMap(type, therapyUuid, mapPath, therapyUuids, name){
+                var indices = getIndexByPath(mapPath);
+                var geneName = $scope.gene.name;
+                var mutationName = $scope.gene.mutations[indices[0]].name;
+                var cancerTypes = $scope.gene.mutations[indices[0]].tumors[indices[1]].cancerTypes;
+                if (type === 'save'){
+                    _.each(therapyUuids, function (drug) {
+                        _.each(cancerTypes, function (cancerType) {
+                            if (cancerType.code != "") {
```
victoria34: Same. Do not use set and remove functions directly. Create corresponding setMap() and removeMap() in firebaseconnector.js for them.

web/yo/app/scripts/controllers/drugs.js
Yizhuzhu: undefined. It works like realtimeString part.

web/yo/app/scripts/services/firebaseconnector.js
```diff
@@ -50,6 +50,21 @@ angular.module('oncokbApp')
...
+        }
+        function removeDrug(uuid){
+            return remove('Drugs/' + uuid);
+        }
+        function addTreatment(path, treatment){
+            return set(path + "/treatments/0", treatment);
+        }
+        function setMap(drug, geneName, mutationName, cancerTypeCode, therapyUuid, name){
+            return set('Map/' + drug + '/' + geneName + '/' + mutationName + '/' + cancerTypeCode + '/' + therapyUuid, name);
+        }
```
victoria34: function setMap(path, name){
            return set('Map/' +path, name);
        }

PR #1483, opened 2019-02-11
Support heme related evidences

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -346,6 +346,16 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
                     new ArrayList<>(relevantAlterations), query));
             }
 
+            // Prognostic summary
+            if (evidenceTypes.contains(EvidenceType.PROGNOSTIC_SUMMARY)) {
```
zhx828: Adding proper prognostic/diagnostic summaries need some work. Will be added in a separate PR.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/LevelsApi.java
```diff
@@ -42,4 +42,22 @@
         method = RequestMethod.GET)
     ResponseEntity<Map<LevelOfEvidence, String>> levelsSensitiveGet();
 
+
+    @ApiOperation(value = "", notes = "Get all prognostic levels.", response = Map.class, tags = {"Levels",})
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "OK", response = Map.class)})
+    @RequestMapping(value = "/levels/prognostic",
```
zhx828: @jjgao just want to make sure the URL path look good to you.

web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -446,6 +446,33 @@ private static void setOncogenic(Gene gene, Alteration alteration, Oncogenicity
         return ret;
     }
 
+    private static void saveTumorLevelSummaries(JSONObject cancerObj, String summaryKey, Gene gene, Set<Alteration> alterations, TumorType oncoTreeType, EvidenceType evidenceType, Integer nestLevel) {
```
zhx828: This is just an abstraction code. Didn't change much.

web/yo/app/scripts/app.js
```diff
@@ -128,28 +128,34 @@ angular.module('oncokbApp').run(
             $rootScope.internal = true;
             $rootScope.meta = {
                 levelsDesc: {
-                    '0': 'FDA-approved drug in this indication irrespective of gene/variant biomarker',
+                    '0': '',
```
zhx828: These levels exist but should not get a description since we never define them in the paper.

zhx828:
@jjgao could you take a look how the endpoints are structured? I'm aware of reorganizing the endpoints base on your request, but before having that, want to make sure anything into production look valid.
@victoria34 could you pay attention to the review module?

Thanks a lot.

PR #1508, opened 2019-03-14
delete EM6

Yizhuzhu:


web/yo/app/scripts/controllers/drugs.js
```diff
@@ -71,7 +75,7 @@ angular.module('oncokbApp')
 
             $scope.removeDrug = function (drug) {
                 if ($scope.drugMap[drug.uuid]) {
-                    var genes = _.map($scope.drugMap[drug.uuid], (gene) => {
+                    var genes = _.map($scope.drugMap[drug.uuid], function(gene){
```
victoria34: I think you can code in this way:
var genes = _.map($scope.drugMap[drug.uuid], 'geneName');
https://lodash.com/docs/#map

web/yo/app/scripts/controllers/drugs.js
```diff
@@ -42,7 +42,11 @@ angular.module('oncokbApp')
 
 
             function hasSameName(newDrugName, uuid) {
-                return _.some(mainUtils.getKeysWithoutFirebasePrefix($scope.drugList), (key) => ($scope.drugList[key].uuid !== uuid && (newDrugName === $scope.drugList[key].drugName || $scope.drugList[key].synonyms !== undefined && $scope.drugList[key].synonyms.indexOf(newDrugName) > -1)) === true);
+                return _.some(mainUtils.getKeysWithoutFirebasePrefix($scope.drugList), function(key){
+                    if(($scope.drugList[key].uuid !== uuid && (newDrugName === $scope.drugList[key].drugName || $scope.drugList[key].synonyms !== undefined && $scope.drugList[key].synonyms.indexOf(newDrugName) > -1)) === true){
```
victoria34: What if it returns `undefined`? What will happen in curation platform?

Yizhuzhu: It will be handled in $scope.saveDrugName. undefined means can't find any.

victoria34: `Undefined` is quite easy to cause errors. Could you just return an empty string?
```
return $scope.drugList[key].uuid !== uuid && (newDrugName === $scope.drugList[key].drugName || $scope.drugList[key].synonyms !== undefined && $scope.drugList[key].synonyms.indexOf(newDrugName) > -1) ? key : '';
```

Yizhuzhu: Sorry, I didn't explain it clearly.  The result returned will be the result of _.some function. According to the Lodash document, "Returns true if any element passes the predicate check, else false".

PR #1514, opened 2019-03-19
Include therapy stats in a Firebase map structure

Yizhuzhu:


web/yo/app/scripts/services/mainutils.js
```diff
@@ -664,6 +664,125 @@ angular.module('oncokbApp')
                 return item.indexOf("$") !== 0;
             });
         }
+        function checkDifferenceBetweenTherapies(oldContent, newContent){
```
zhx828: If you want to make this a generic method and export, you probably should accept two arrays instead of two string. I don't see much of the use case here by exporting the method if the input needs to be therapy string format.

web/yo/app/scripts/services/mainutils.js
```diff
@@ -703,6 +822,11 @@ angular.module('oncokbApp')
             therapyStrToArr: therapyStrToArr,
             getDrugNameByUuids: getDrugNameByUuids,
             drugUuidtoName: drugUuidtoName,
-            getKeysWithoutFirebasePrefix: getKeysWithoutFirebasePrefix
+            getKeysWithoutFirebasePrefix: getKeysWithoutFirebasePrefix,
```
zhx828: You need to consider whether you want to export all following methods. If they are not potentially used, they should not be exported. And since they all in mainUtils, they need to be generic enough in order to have the name like `changeMapWhenAdd`. You could also consider leaving these methods to DrugMapUtils for less strict naming convention.

web/yo/app/scripts/services/mainutils.js
```diff
@@ -664,6 +664,125 @@ angular.module('oncokbApp')
...
+                    }
+                    else {
+                        drugArray = drugs;
+                    }
+                    var therapyObject = {
+                        'name': content,
+                        'status': 'latest'
+                    };
+                    var mapPath = '';
+                    var chainDrug = $q.when();
```
zhx828: If we enable typescript, the way you manipulating chainDrug is not acceptable. Rather than keep attaching to the chainDrug, why not just create an array and keep executing until the array is empty?

web/yo/app/scripts/services/mainutils.js
```diff
@@ -664,6 +664,125 @@ angular.module('oncokbApp')
...
+                    _.forEach(drugArray, function (drug) {
```
zhx828: Also, you need to handle error.

web/yo/app/scripts/services/mainutils.js
```diff
@@ -664,6 +664,125 @@ angular.module('oncokbApp')
...
+                case 'mutation':
+                    break;
+                case 'tumor':
+                    break;
+                case 'treatment':
+                    //Functions here should be optimized.
+                    _.forEach(drugArray, function(drug){
+                        var chainDrug = $q.when();
+                        var mapPath = '';
+                        chainDrug = chainDrug.then(function(){
```
zhx828: Please update base on the comments above.

web/yo/app/scripts/services/mainutils.js
```diff
@@ -664,6 +664,125 @@ angular.module('oncokbApp')
...
+            switch (dataType){
+                case 'mutation':
+                    break;
+                case 'tumor':
+                    break;
+                case 'treatment':
+                    _.forEach(drugArray, function(drug){
+                        var mapPath = '';
+                        var chainDrug = $q.when();
+                        chainDrug = chainDrug.then(function(){
```
zhx828: Same here.

web/yo/app/scripts/services/mainutils.js
zhx828: Consider creating an util to run through the promise array.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1522,52 +1521,172 @@ angular.module('oncokbApp')
...
+                switch (dataType){
+                    case 'mutation':
+                        break;
+                    case 'tumor':
+                        break;
+                    case 'treatment':
+                        var mapPath = '';
+                        var chainDrug = $q.when();
+                        _.forEach(drugArray, function (drug) {
+                            chainDrug = chainDrug.then(function(){
```
zhx828: Please take a look the comment about the chainDrug in `firebaseconnector.js`

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1522,52 +1521,172 @@ angular.module('oncokbApp')
...
+                        break;
+                }
+            }
+
+            function removeMapWithoutValidatingStatus (geneName, mutationUuid, cancerTypeUuid, therapyUuid, therapyUuids){
+                var mapPath = '';
+                _.each(therapyUuids, function (drug) {
+                    mapPath = drug + '/' + geneName + '/' + mutationUuid + '/cancerTypes/' + cancerTypeUuid + '/' + therapyUuid;
+                    firebaseConnector.removeMap(mapPath).then(function(){
+                        firebaseConnector.ref('Map/' + drug + '/' + geneName + '/' + mutationUuid + '/cancerTypes').once('value', function (doc) {
```
zhx828: Can we put these concatenating to into string util?

zhx828: Something calls like `getCancerTypesPath`...

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1522,52 +1521,172 @@ angular.module('oncokbApp')
...
-                            case 'treatment':
-                                var oldTherapyUuids = _.flatten(mainUtils.therapyStrToArr(oldContent));
-                                changeMap('remove', treatment.name_uuid, mapPath, oldTherapyUuids);
-                                changeMap('save', treatment.name_uuid, mapPath, TherapyUuids, treatment.name);
-                                break;
+                }
+            }
+
+            function updateDrugMapWhenAcceptDeletionOrRejectAddition(dataType, mutation, tumor, treatment, mapInfo){
+                switch (dataType) {
```
zhx828: Technically, all switch should attach a default handler, and leave a comment there for all scenarios that not qualify previous cases.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1982,23 +2102,26 @@ angular.module('oncokbApp')
                         removeModel({ type: type, path: path, uuids: uuids });
                     } else {
                         if (type === 'tumor') {
-                            obj.cancerTypes_review = {
-                                updatedBy: $rootScope.me.name,
-                                updateTime: new Date().getTime(),
-                                removed: true
-                            }
+                            // Do not delete lastReviewed when curators delete a treatment.
```
zhx828: Why changes in PR #1521 also presents in here?

web/yo/app/scripts/controllers/tools.js
```diff
@@ -600,4 +600,14 @@ angular.module('oncokbApp')
                     $scope.selectedTypeCheckboxes.push(checkbox);
                 }
             };
+            $scope.testInternal = function () {
```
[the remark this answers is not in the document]
Yizhuzhu: My bad. It should be deleted. Internal network related code will be shown in next pr.

web/yo/app/scripts/services/mainutils.js
Yizhuzhu: This function is used in three different places. If I change it to arrays,  I should preprocess Strings in these three different places. And 4 more lines of similar codes are needed.

zhx828: Don't export the function if that's the case since it's highly customized.

zhx828: You can export a generic one and create a wrapper on top for local usage.

web/yo/app/views/tools.html
```diff
@@ -1,3 +1,4 @@
+<div ng-init="testInternal()" ng-show="internalNetwrok === false">Reminder: Please use internal network to use all functions.</div>
```
zhx828: Is the function still there? `testInternal `

web/yo/app/scripts/controllers/gene.js
```diff
@@ -1340,17 +1340,15 @@ angular.module('oncokbApp')
                             { reviewObj: treatment.description_review, uuid: treatment.description_uuid }], treatment.name_uuid);
                         break;
                     case 'MUTATION_NAME_CHANGE':
-                        updateDrugMap('name', 'mutation', mutation, tumor, treatment, mutation.name_review, mapPath);
+                        //$scope.updateDrugMap('accept', 'name', 'mutation', mutation, tumor, treatment, mutation.name_review);
```
zhx828: Don't needed anymore?

web/yo/app/scripts/services/drugmaputils.js
```diff
@@ -0,0 +1,176 @@
...
+                            })
+                        })
+                    });
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        function getMapPath(drug, geneName, mutationUuid, cancerTypeUuid, therapyUuid){
```
zhx828: This needs to be done more thoughtfully. The purpose of this function is getting the path to the therapyInMap.
Also please take a look all other places where you generate the path. These should all be done in a Util. Maybe FirebasePathUtils. For instance, in function `changeTherapyNameInMapWithValidatingStatus`

web/yo/app/scripts/directives/addtherapy.js
```diff
@@ -1,17 +1,27 @@
...
-    .directive('addTherapy', function ($rootScope, DatabaseConnector, _, $q, FirebaseModel, firebaseConnector, mainUtils, $window) {
+    .directive('addTherapy', function ($rootScope, DatabaseConnector, _, $q, FirebaseModel, firebaseConnector, drugMapUtils, $window) {
         return {
             templateUrl: 'views/addTherapy.html',
             restrict: 'E',
+            scope:{
+                modifyName: '=',
+                tiRef: '=',
+                treatmentRef: '=',
+                saveCallbackInGene: '&addNewTherapy',
```
zhx828: Why do you need to differentiate where the component is used?

web/yo/app/scripts/services/drugmaputils.js
```diff
@@ -0,0 +1,176 @@
...
+            return drug + '/' + geneName + '/' + mutationUuid + '/cancerTypes/' + cancerTypeUuid + '/' + therapyUuid;
+        }
+
+        return {
+            therapyStrToArr: therapyStrToArr,
+            drugUuidtoName: drugUuidtoName,
+            getKeysWithoutFirebasePrefix: getKeysWithoutFirebasePrefix,
+            checkDifferenceBetweenTherapies: checkDifferenceBetweenTherapies,
```
zhx828: Is this function used anywhere except in this file?

web/yo/app/scripts/directives/addtherapy.js
Yizhuzhu: These two functions are defined in two controllers. One is Gene Controller, the other one is ModifyTherapy controller. I couldn't define both as saveCallback.

zhx828: Back to previous discussion, the component should not aware of where it will be used. You don't need to define `add-new-therapy` property for GeneController and `save-modification` for ModifyController. They should all be `save-callback`.

web/yo/app/scripts/services/drugmaputils.js
Yizhuzhu: checkDifferenceBetweenTherapies is used in gene.js.

PR #1521, opened 2019-03-21
fixRollbackBug

Yizhuzhu:
Rollback to last reviewed content, not the latest content.


web/yo/app/scripts/directives/realtimestring.js
```diff
@@ -176,6 +176,14 @@ angular.module('oncokbApp')
                 $scope.setTrackSignal = function() {
                     mainUtils.updateMovingFlag(false);
                 };
+                $scope.getMutationName = function(key, oldKey, uuid){
+                    if(mainUtils.processedInReview('remove', uuid) && oldKey){
```
victoria34: Do you need to check whether `reviewmode === true` in here?

Yizhuzhu: I don't think so. This function will only be triggered in review mode.

victoria34: What if 2 curators work on the same gene?

PR #1535, opened 2019-03-28
Fixes related to Drugs

zhx828:
- Move drugList to rootScope
- Update DB after changing the preferred name
- Let Save Genes and Update gene working properly

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
@@ -4,10 +4,12 @@
  */
 package org.mskcc.cbio.oncokb.controller;
 
+import com.google.gson.JsonArray;
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.mskcc.cbio.oncokb.apiModels.NCITDrug;
 import org.mskcc.cbio.oncokb.bo.*;
 import org.mskcc.cbio.oncokb.model.*;
 import org.mskcc.cbio.oncokb.model.oncotree.TumorType;
@@ -574,17 +576,17 @@ public class DriveAnnotationParser {
 
         // specific evidence
         DrugBo drugBo = ApplicationContextSingleton.getDrugBo();
-        JSONArray drugsArray = implicationObj.has("treatments") ? implicationObj.getJSONArray("treatments") : new JSONArray();
+        JSONArray treatmentsArray = implicationObj.has("treatments") ? implicationObj.getJSONArray("treatments") : new JSONArray();
         int priorityCount = 1;
-        for (int i = 0; i < drugsArray.length(); i++) {
-            JSONObject drugObj = drugsArray.getJSONObject(i);
-            if (!drugObj.has("name") || drugObj.getString("name").trim().isEmpty()) {
+        for (int i = 0; i < treatmentsArray.length(); i++) {
+            JSONObject drugObj = treatmentsArray.getJSONObject(i);
+            if (!drugObj.has("name") || drugObj.getJSONArray("name").length() == 0) {
                 System.out.println(spaceStrByNestLevel(nestLevel + 1) + "Drug does not have name, skip... " + drugObj.toString());
                 continue;
             }
 
-            String drugNameStr = drugObj.getString("name").trim();
-            System.out.println(spaceStrByNestLevel(nestLevel + 1) + "Drug(s): " + drugNameStr);
+            JSONArray therapiesArray = drugObj.getJSONArray("name");
+            System.out.println(spaceStrByNestLevel(nestLevel + 1) + "Drug(s): " + therapiesArray.length());
 
             Set<Date> lastEditDates = new HashSet<>();
             addDateToSetFromObject(lastEditDates, drugObj, "name_review");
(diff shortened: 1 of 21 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -607,24 +609,45 @@ private static void parseTherapeuticImplications(Gene gene, Set<Alteration> alte
                 addDateToSetFromObject(lastEditDates, drugObj, "indication_review");
             }
 
-            String[] drugTxts = drugNameStr.replaceAll("(\\([^\\)]*\\))|(\\[[^\\]]*\\])", "").split(",");
-
             List<Treatment> treatments = new ArrayList<>();
-            for (int j = 0; j < drugTxts.length; j++) {
-                String[] drugNames = drugTxts[j].split(" ?\\+ ?");
+            for (int j = 0; j < therapiesArray.length(); j++) {
```
Yizhuzhu: j < therapiesArray.length();  Brackets here might be unnecessary.

web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -607,24 +609,45 @@ private static void parseTherapeuticImplications(Gene gene, Set<Alteration> alte
...
+                JSONArray drugsArray = therapiesArray.getJSONArray(j);
 
                 List<Drug> drugs = new ArrayList<>();
-                for (String drugName : drugNames) {
-                    drugName = drugName.trim();
-                    Drug drug = drugBo.guessUnambiguousDrug(drugName);
+                for (int k = 0; k < drugsArray.length(); k++) {
```
Yizhuzhu: Same here about brackets.

web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -574,17 +576,17 @@ private static void parseTherapeuticImplications(Gene gene, Set<Alteration> alte
 
         // specific evidence
         DrugBo drugBo = ApplicationContextSingleton.getDrugBo();
-        JSONArray drugsArray = implicationObj.has("treatments") ? implicationObj.getJSONArray("treatments") : new JSONArray();
+        JSONArray treatmentsArray = implicationObj.has("treatments") ? implicationObj.getJSONArray("treatments") : new JSONArray();
         int priorityCount = 1;
-        for (int i = 0; i < drugsArray.length(); i++) {
-            JSONObject drugObj = drugsArray.getJSONObject(i);
-            if (!drugObj.has("name") || drugObj.getString("name").trim().isEmpty()) {
+        for (int i = 0; i < treatmentsArray.length(); i++) {
```
Yizhuzhu: i < treatmentsArray.length(). Brackets here might be unnecessary.
For array, generally use length. And for String object, use length()?

web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -574,17 +576,17 @@ private static void parseTherapeuticImplications(Gene gene, Set<Alteration> alte
...
+            JSONObject drugObj = treatmentsArray.getJSONObject(i);
+            if (!drugObj.has("name") || drugObj.getJSONArray("name").length() == 0) {
```
Yizhuzhu: drugObj.getJSONArray("name").length()     ()here

web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
```diff
@@ -574,17 +576,17 @@ private static void parseTherapeuticImplications(Gene gene, Set<Alteration> alte
...
                 System.out.println(spaceStrByNestLevel(nestLevel + 1) + "Drug does not have name, skip... " + drugObj.toString());
                 continue;
             }
 
-            String drugNameStr = drugObj.getString("name").trim();
-            System.out.println(spaceStrByNestLevel(nestLevel + 1) + "Drug(s): " + drugNameStr);
+            JSONArray therapiesArray = drugObj.getJSONArray("name");
+            System.out.println(spaceStrByNestLevel(nestLevel + 1) + "Drug(s): " + therapiesArray.length());
```
Yizhuzhu: ()

web/src/main/java/org/mskcc/cbio/oncokb/controller/DriveAnnotationParser.java
zhx828: This is JAVA, and that's a method not property.

Yizhuzhu: I found the following words on https://www.geeksforgeeks.org/length-vs-length-java/
length vs length() in Java
array.length : length is a final variable applicable for arrays. With the help of length variable, we can obtain the size of the array.
string.length() : length() method is a final variable which is applicable for string objects. length() method returns the number of characters presents in the string.

I am not sure. Just a reminder. Sorry for it.

zhx828: @Yizhuzhu it depends on how the class is designed. It could be a public accessible property or a method. In these few places, JSONArray only implements the method instead of expose the property. I would not `worry` that much since the IDE will usually give you suggestions and it would not be compiled if that's actually the problem.

victoria34:
@zhx828  I approved the PR for all backend code.
@Yizhuzhu I think you still need to review front-end code.

PR #1552, opened 2019-04-04
Allow admin manually release gene

zhx828:
Option to Release/Remove genes(Release is for admin only, remove is for developers only):
Screen Shot 2019-04-04 at 5 42 44 PM

Status check for releasing gene:
Screen Shot 2019-04-04 at 5 38 18 PM

### DIFF
--- a/web/yo/app/scripts/controllers/gene.js
+++ b/web/yo/app/scripts/controllers/gene.js
@@ -409,11 +409,6 @@ angular.module('oncokbApp')
                     return {'margin-top': '20px'};
                 }
             };
-            $scope.removeGeneFromDB = function() {
-                DatabaseConnector.removeGeneFromDB($scope.gene.name, function(data) {
-                    console.log(data);
-                });
-            };
             /**
              * This function is used to find the most recent update from a section change. e.g. There are 3 items under Prognostic section, and they might get changed at very different time.
              * And we will find the one changed most recently and store them in ReviewResource.mostRecent mapping, so it could be shared across directives and controllers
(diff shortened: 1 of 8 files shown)


web/yo/app/scripts/controllers/gene.js
```diff
@@ -454,11 +454,10 @@ angular.module('oncokbApp')
                 return result;
             }
             var userNames = [];
-            function prepareReviewItems() {
```
zhx828: This function is kinda complicated, we might need to do some refactoring for proper usage. For now, I'm just replace the scope status to function-wide variable. These three are not really used across genes page. The logic of `sectionUUIDs` seems not really straight forward since it has been used in another two methods. So I'm gonna just assign it to global sectionUUIDs at the place calls getReviewInfo.

web/yo/app/scripts/controllers/gene.js
```diff
@@ -3302,6 +3435,55 @@ angular.module('oncokbApp')
         }]
     )
 
+    .controller('ReleaseGeneDialogCtrl', function($scope, $modalInstance, data) {
+        $scope.statusChecks = data.statusChecks || [];
+        $scope.error='';
+
+        _.each($scope.statusChecks, function(statusCheck) {
+            if (!statusCheck.status) {
```
zhx828: The logic here is
- if the statusCheck item already has a status which means the logic is decided outside of the controller. Nothing really needs to be done here.
- If the promise is not given, it should be an optional check. So assign a warning status.
- Depends on the promise status to assign the statusCheck status

zhx828:
@Yizhuzhu @victoria34 while working on this PR, I realize the bootstrap comes with lots of text styles. So for warning/error/information messages, you do not need to create your own color style on top. Just assign proper color class name. Please see https://getbootstrap.com/docs/3.3/css/

web/yo/app/scripts/controllers/gene.js
```diff
@@ -454,11 +454,10 @@ angular.module('oncokbApp')
                 return result;
             }
             var userNames = [];
-            function prepareReviewItems() {
-                resetReviewResources();
-                $scope.sectionUUIDs = []; // sectionUUIDs is used to store uuid per section.
-                $scope.status.hasReviewContent = false;
-                $scope.status.mutationChanged = false;
+            function getReviewInfo() {
+                var sectionUUIDs_ = [];
```
victoria34: where is `resetReviewResources();`?

PR #1555, opened 2019-04-05
rootScope.drugList & Refining therapy result

Yizhuzhu:


Yizhuzhu:
To handle the following issues:
1. The curator goes to gene page or drug page directly without logging in. The rootScope.drugList will be undefined.
2. Extra commas in TherapyResult if curators let one input box empty
In addition,
1. To hide Therapy 1's delete button
2. Keep save button unable when the curator delete one empty input box.

PR #1556, opened 2019-04-08
Validate PMIDs

### LINKED ISSUE #1540: Curation Assistant
- [x] give warning message when a PMID is not available. PR#1556

victoria34:
Fix #1540 
Screen Shot 2019-04-08 at 12 16 47 PM


web/yo/app/scripts/services/pubUtils.js
```diff
@@ -0,0 +1,43 @@
...
+ * This factory is a cache service used for storing validated PMIDs.
+ */
+angular.module('oncokbApp')
+    .factory('pubUtils', function() {
+        var validatedPubs = {};
+
+        function clear() {
+            validatedPubs = {};
+        }
+        function get(pubs) {
```
zhx828: pubs should just be a string array.

web/yo/app/scripts/services/pubUtils.js
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @ngdoc factory
+ * @name oncokbApp.pubUtils
+ * @author Jing Su
+ * @description
+ * This factory is a cache service used for storing validated PMIDs.
+ */
+angular.module('oncokbApp')
+    .factory('pubUtils', function() {
+        var validatedPubs = {};
```
zhx828: Do we need to cache the full content of the PMID?

web/yo/app/scripts/directives/pubiframe.js
```diff
@@ -22,16 +22,21 @@ angular.module('oncokbApp')
...
-                    //     }, function (error) {
-                    //         console.log('Error happened', error);
-                    //     });
-                    // }
+                    if (pubs.length > 0) {
+                        var pubsCache = pubUtils.get(pubs);
+                        if (pubsCache.notValidatedPuds.length > 0) {
+                            FindRegex.validation(pubsCache.notValidatedPuds).then(function(result) {
+                                pubUtils.set(result);
+                                scope.pubs = _.union(pubsCache.validatedPuds, result);
```
zhx828: There would not be any duplicate in validatedPuds from result.

web/yo/app/scripts/services/pubUtils.js
victoria34: I need to store the whole pub structure for not validated PMIDs later.

web/yo/app/scripts/directives/pubiframe.js
victoria34: Right. I'll use _.concat() in here.

web/yo/app/scripts/services/pubUtils.js
victoria34: The full content of PMID are `type, id and valid`. If you think the data is too much, I can only store `valid`.

web/yo/app/scripts/directives/pubiframe.js
zhx828: Why there are duplications?

web/yo/app/scripts/services/pubUtils.js
zhx828: Oh right, you also want to cache . the invalid as well. Cool, thanks for the info.

web/yo/app/scripts/directives/pubiframe.js
victoria34: Just in case curators will type same PMID twice.

web/yo/app/scripts/services/pubCache.js
```diff
@@ -0,0 +1,54 @@
...
+            };
+        }
+        function set(pubs) {
+            _.each(pubs, function(pub){
+                validatedPubs[pub.id] = pub;
+            });
+        }
+        function validatePub(pubs) {
+            var deferred = $q.defer();
+            FindRegex.validation(pubs).then(function(result) {
```
zhx828: Don't you want to utilize the cache first for known pmids?

victoria34: Yes, I used the get() to load known pmids first and also get pmids that are not validated. Then I'll use validatePub() to validate those unknown pmids.

zhx828: Oh I see. Why not just include that get logic in the validatePub?

victoria34: I want to separate them to keep each function minimal.

zhx828: I would not worry about make the function minimal particularly  for this function. It's a util function anyway. But up to you.

PR #1570, opened 2019-04-22
Update when out of internal network & bug fixes

zhx828:
- Make sure the synonyms is always an array
- Do not allow changing preferred name
- Update tabs once the internal status is determined
- Use curation private for search drug
- Add a loading bar when changing the preferred name

### DIFF
--- a/web/yo/app/scripts/controllers/nav.js
+++ b/web/yo/app/scripts/controllers/nav.js
@@ -23,12 +23,13 @@ angular.module('oncokbApp')
             filterTabs.push('genes');
             filterTabs.push('queues');
             filterTabs.push('therapies');
-            if ($rootScope.me.admin) {
+            filterTabs.push('tools');
+            if ($rootScope.me && $rootScope.me.admin) {
                 filterTabs = _.union(filterTabs, ['variant', 'tools', 'feedback']);
             }
 
             if (!$rootScope.internal) {
-                filterTabs = _.intersection(filterTabs, ['genes', 'queues', 'feedback']);
+                filterTabs = _.intersection(filterTabs, ['genes', 'queues', 'feedback', 'therapies', 'tools']);
             }
 
             $scope.tabs = filterTabs.map(function(tabKey) {
@@ -40,18 +41,6 @@ angular.module('oncokbApp')
                 delete window.localStorage.geneName;
             }
             window.localStorage.tab = key;
-        }
-
-        function testInternal() {
-            var defer = $q.defer();
-            DatabaseConnector.testAccess(function() {
-                $rootScope.internal = true;
-                defer.resolve();
-            }, function(data, status, headers, config) {
-                $rootScope.internal = false;
-                defer.resolve();
-            });
-            return defer.promise;
         }
         $firebaseAuth().$onAuthStateChanged(function(firebaseUser) {
             if (firebaseUser) {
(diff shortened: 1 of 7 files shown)


web/yo/app/scripts/controllers/nav.js
```diff
@@ -23,12 +23,13 @@ angular.module('oncokbApp')
             filterTabs.push('genes');
             filterTabs.push('queues');
             filterTabs.push('therapies');
-            if ($rootScope.me.admin) {
+            filterTabs.push('tools');
+            if ($rootScope.me && $rootScope.me.admin) {
                 filterTabs = _.union(filterTabs, ['variant', 'tools', 'feedback']);
```
victoria34: Are we supposed to remove `tools` in here?

web/yo/app/scripts/controllers/nav.js
```diff
@@ -23,12 +23,13 @@ angular.module('oncokbApp')
             filterTabs.push('genes');
```
victoria34: Could we just write like this?
```
var filterTabs = ['genes', 'queues', 'therapies', 'tools'];  
```

web/yo/app/scripts/controllers/nav.js
```diff
@@ -23,12 +23,13 @@ angular.module('oncokbApp')
...
             filterTabs.push('therapies');
-            if ($rootScope.me.admin) {
+            filterTabs.push('tools');
+            if ($rootScope.me && $rootScope.me.admin) {
                 filterTabs = _.union(filterTabs, ['variant', 'tools', 'feedback']);
             }
 
             if (!$rootScope.internal) {
-                filterTabs = _.intersection(filterTabs, ['genes', 'queues', 'feedback']);
+                filterTabs = _.intersection(filterTabs, ['genes', 'queues', 'feedback', 'therapies', 'tools']);
```
victoria34: Does it mean we can only see the `Variant Annotation` tab when we are in external network?

web/yo/app/scripts/controllers/nav.js
zhx828: Tools is only available for admins

web/yo/app/scripts/controllers/nav.js
zhx828: This is previous setting. I think it's ok.

web/yo/app/scripts/controllers/nav.js
zhx828: This is intersection when user is in external. Variant Annotation should only available internal

Yizhuzhu:
I downloaded it and did a test. I have two concerns.
1. Change network from internal to public website. Nothing changed in nav bar and tools page.
2. Changing drug preferred name. When I clicked the check button to save, nothing responded during two seconds and then the page flashed. I guess it was loading page. Some words on the page, but it flashed too quickly. It's hard to read it.

zhx828:
@Yizhuzhu for your concerns:
1: I don't think in general we have the solution when toggle from internal to external or the other way. User has to do refresh the page.
2: The reason I added the loading status is because changing preferred name is very expensive. It usually takes about one minute or even longer. The server side needs to do full recaching. So I don't think it will flash. But could you explain a bit more about `nothing responded`?

web/yo/app/scripts/controllers/nav.js
victoria34: So we should remove line 26 right?

Yizhuzhu:
@zhx828 After clicking save button, I need to wait about 2 seconds, then the page flashes in about 1 second. Then it goes back to drug page. The time may not be exact. However, flashing time is less than waiting time.

web/yo/app/scripts/controllers/nav.js
zhx828: Ah yes, let me do that.

zhx828:
@Yizhuzhu there is no so called flashing time is less than waiting time. The loading bar will be shown for quite long(see my comment above). I update the loading status right after the saveDrugName is called. It may just due to the lifecycle. No sure what I can do here. But I'm happy to invest in more time on this issue later. But I don't think it's a priority at the moment.

PR #1577, opened 2019-04-25
Bug fixs

zhx828:


core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -253,10 +254,87 @@ private static void cacheAllGenes() {
...
+                    cancerGene.setOncokbAnnotated(true);
+                    cancerGene.setOncogene(gene.getOncogene());
+                    cancerGene.setTSG(gene.getTSG());
+                }
+                int occurence = NumberUtils.isNumber(items[2].trim()) ? Integer.parseInt(items[2].trim()) : 0;
+                if (cancerGene.getOncokbAnnotated()) {
+                    occurence++;
+                }
+                cancerGene.setOccurrenceCount(occurence);
+                cancerGene.setmSKImpact(items[3].trim().equals("1"));
```
victoria34: What does `equals("1")` mean in here?  Do you mean the length of element after trim()?

zhx828: This is the actual value from the file. In the file, this column, 1 means exists

PR #1584, opened 2019-05-06
update

Yizhuzhu:
1. Add Therapy component's CSS.
2. Return drugs in order when adding therapy.

### DIFF
--- a/web/yo/app/scripts/directives/addtherapy.js
+++ b/web/yo/app/scripts/directives/addtherapy.js
@@ -77,10 +77,11 @@ angular.module('oncokbApp')
 
                 $scope.loadDrugs = function($query) {
                     $scope.addTherapyError = false;
-                    return drugs.filter(function (drug) {
-                        var lowerCaseQuery = $query.toLowerCase();
-                        return drug.drugName.toLowerCase().indexOf(lowerCaseQuery) != -1 || (!drug.synonyms ? false : drug.synonyms.join(',').toLowerCase().indexOf(lowerCaseQuery) != -1);
+                    var lowerCaseQuery = $query.toLowerCase();
+                    var filteredDrugs = drugs.filter(function (drug){
+                        return drug.drugName.toLowerCase().indexOf(lowerCaseQuery) !== -1 || (!drug.synonyms ? false : drug.synonyms.join(',').toLowerCase().indexOf(lowerCaseQuery) !== -1);
                     })
+                    return getDrugsInOrder(lowerCaseQuery, filteredDrugs);
                 };
 
                 $scope.addDruginTherapy = function (uuid, index) {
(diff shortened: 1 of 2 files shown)


web/yo/app/scripts/directives/addtherapy.js
```diff
@@ -77,10 +77,15 @@ angular.module('oncokbApp')
 
                 $scope.loadDrugs = function($query) {
                     $scope.addTherapyError = false;
-                    return drugs.filter(function (drug) {
-                        var lowerCaseQuery = $query.toLowerCase();
-                        return drug.drugName.toLowerCase().indexOf(lowerCaseQuery) != -1 || (!drug.synonyms ? false : drug.synonyms.join(',').toLowerCase().indexOf(lowerCaseQuery) != -1);
-                    })
+                    var lowerCaseQuery = $query.toLowerCase();
+                    var result = [];
+                    result = drugs.filter(function (drug){
```
zhx828: Why twice?

Yizhuzhu: First: return drugs whose drug name contain the query.
Second: return drugs whose synonyms contain the query.
To make drug list in order.

zhx828: I would instead considering just introduce a sorting function. `.filter` is not for this purpose, and you also need to do a uniq and concat. Please also consider then a drug name starts with the query should have a higher priority then the drug name has the query.

web/yo/app/scripts/directives/addtherapy.js
```diff
@@ -97,6 +98,36 @@ angular.module('oncokbApp')
                     addTherapy(index);
                 };
 
+                function getDrugsInOrder(lowerCaseQuery) {
+                    _.forEach(drugs, function(drug){
```
zhx828: It's minor, but I would rather use _.map to assign the weight. It's more elegant than constantly returning.

web/yo/app/scripts/directives/addtherapy.js
```diff
@@ -77,10 +77,11 @@ angular.module('oncokbApp')
 
                 $scope.loadDrugs = function($query) {
                     $scope.addTherapyError = false;
+                    var lowerCaseQuery = $query.toLowerCase();
+                    getDrugsInOrder(lowerCaseQuery);
                     return drugs.filter(function (drug) {
```
zhx828: And I would rather do a filter then a sort to make it more efficient. Also you don't really need to sign a weight on items you don't want to include.

web/yo/app/scripts/directives/addtherapy.js
```diff
@@ -97,6 +98,36 @@ angular.module('oncokbApp')
                     addTherapy(index);
                 };
 
+                function getDrugsInOrder(lowerCaseQuery) {
+                    _.forEach(drugs, function(drug){
+                        if(drug.drugName.toLowerCase()===lowerCaseQuery){
```
zhx828: drugName should be lowercased  at the top to avoid repeat converting

PR #1608, opened 2019-05-20
Remove blank space character

victoria34:


web/yo/app/scripts/directives/realtimestring.js
```diff
@@ -53,6 +53,15 @@ angular.module('oncokbApp')
                         });
                     }
                     scope.$watch('data[key]', function (n, o) {
+                        // Remove blank space character
+                        if (!_.isUndefined(n) && !_.isUndefined(o) && (n.indexOf('&nbsp') > -1 || o.indexOf('&nbsp') > -1)) {
```
zhx828: I think one simple `_.replace` would work?

victoria34: You want to replace `&nbsp` directly?

zhx828: actually `&nbsp;`

zhx828: And this is maybe for later, but the solution would probably just a temporary solution because there may be more HTML reserved characters. But haven't seen any yet.

victoria34: The issued you showed to me is not caused by copy and paste. I think curators type blank space accidentally and approve the change so you can see `&nbsp;` on public website. In that case, we don't need to worry about HTML tags reserved scenario because curators cannot type html. We only need to check if `&nbsp` reserved in typed field and clean it.

zhx828: If it's not introduced in copy/paste, how are the characters introduced?

victoria34: The issue shows `&nbsp` in mutation name. I tried to copy and paste mutation name with blank space and the getString function work, so I guess curators may press blank space during typing and they didn't notice that.

web/yo/app/scripts/services/mainutils.js
```diff
@@ -644,6 +644,34 @@ angular.module('oncokbApp')
...
+            var _string = tmp.textContent || tmp.innerText || S(string).stripTags().s;
+            string = S(_string).collapseWhitespace().s;
+            string = string.replace(/<!--.*-->/g, '');
+
+            // Convert the html fragment again to trim the reserved text.
+            var tmp = window.document.createElement('DIV');
+            tmp.innerHTML = string;
+
+            var _string = tmp.textContent || tmp.innerText || S(string).stripTags().s;
+            string = S(_string).decodeHTMLEntities().s;
```
victoria34: decodeHTMLEntities() can help to convert html symbol to characters(like `&nbsp;` -> ` `) but remove html tags, so I inserted decodeHTMLEntities() in getString() and didn't change other code.

PR #1612, opened 2019-05-24
TestUtils & sendEmails

Yizhuzhu:
Test DrugMapUtils & FirebasePathUtils

### DIFF
--- a/web/yo/app/scripts/services/drugmaputils.js
+++ b/web/yo/app/scripts/services/drugmaputils.js
@@ -10,32 +10,52 @@
 angular.module('oncokbApp')
     .factory('drugMapUtils', function(OncoKB, _, $q, DatabaseConnector, $rootScope, ReviewResource, S, UUIDjs, $routeParams, firebaseConnector, firebasePathUtils) {
 
-        function therapyStrToArr(key){
-            return key.split(",").map(function(element){
-                return element.trim().split(" + ")
-            });
+        function therapyStrToArr(key) {
+            if (key) {
+                return key.split(",").map(function (element) {
+                    return element.trim().split(" + ")
+                });
+            } else {
+                return [];
+            }
         }
-        function drugUuidtoName(key, drugList){
-            if (key != undefined){
+
+        function drugUuidtoName(key, drugList) {
+            if (key != undefined) {
                 var keys = therapyStrToArr(key);
                 return getDrugNameByUuids(keys, drugList);
+            } else {
+                return key;
             }
         }
-        function drugUuidtoDrug(key, drugList){
-            if (key != undefined){
+
+        function drugUuidtoDrug(key, drugList) {
+            if (key != undefined) {
                 var keys = therapyStrToArr(key);
                 return getDrugsByUuids(keys, drugList);
+            } else {
+                return {};
             }
         }
-        function getKeysWithoutFirebasePrefix(array){
-            return _.keys(array).filter(function (item) {
-                return item.indexOf("$") !== 0;
-            });
+
+        function getKeysWithoutFirebasePrefix(array) {
+            if (array) {
+                return _.keys(array).filter(function (item) {
+                    return item.indexOf("$") !== 0;
+                });
+            } else {
+                return array;
+            }
         }
-        function checkDifferenceBetweenTherapies(oldContent, newContent){
-            var oldArray = _.flatten(therapyStrToArr(oldContent));
-            var newArray = _.flatten(therapyStrToArr(newContent));
-            return checkDifferenceBetweenTwoArrays(oldArray, newArray);
+
+        function checkDifferenceBetweenTherapies(oldContent, newContent) {
+            if (oldContent || newContent) {
+                var oldArray = _.flatten(therapyStrToArr(oldContent));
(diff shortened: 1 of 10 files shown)


web/yo/app/scripts/services/user.js
```diff
@@ -11,7 +11,7 @@
  * gmail.com will be used as standard email format, googlemail address will be converted to gmail
  */
 angular.module('oncokbApp')
-    .service('user', function user($routeParams, $q, $firebaseAuth, $firebaseObject, $rootScope, mainUtils, firebaseConnector) {
+    .service('user', function user($routeParams, $q, $firebaseAuth, $firebaseObject, $rootScope, mainUtils, firebaseConnector, dialogs) {
```
zhx828: No change needed?

web/yo/app/scripts/services/drugmaputils.spec.js
```diff
@@ -0,0 +1,92 @@
...
+            "description" : "",
+            "drugName" : "Ponatinib",
+            "ncitCode" : "C95777",
+            "ncitName" : "Ponatinib",
+            "synonyms" : [ "PONATINIB", "AP-24534", "AP24534", "Benzamide, 3-(2-Imidazo(1,2-B)Pyridazin-3-Ylethynyl)-4-Methyl-N-(4-((4-Methyl-1- Piperazinyl)Methyl)-3-(Trifluoromethyl)Phenyl)" ],
+            "uuid" : "122b8921-3cd6-4b57-bdc8-5d39cc5465a1"
+        }
+    };
+    it('therapyStrToArr should change therapyName from String to Array', function () {
+        expect(drugMapUtilsFactory.therapyStrToArr('Encorafenib + Pembrolizumab, MDM2 Antagonist RO5045337, MDM2 Antagonist RO5045337 + Nilotinib, Ponatinib')).toEqual([['Encorafenib', 'Pembrolizumab'], ['MDM2 Antagonist RO5045337'], ['MDM2 Antagonist RO5045337', 'Nilotinib'], ['Ponatinib']]);
```
zhx828: Could you also do some extreme tests, like  if the input is null, undefined or empt string? You could also try to test some format cases. For instance `Encorafenib +Pembrolizumab`, `Encorafenib+ Pembrolizumab`, ` Encorafenib+Pembrolizumab ` is supposed to be `([['Encorafenib', 'Pembrolizumab']]`

web/yo/app/scripts/services/user.js
Yizhuzhu: dialogs has been added to master branch. I have no idea why there is difference here. dialogs is needed.

web/yo/app/scripts/services/drugmaputils.spec.js
Yizhuzhu: For therapyStrToArray function, the input can't be null or undefined. It has been checked before triggering this function. And removing duplicates is not needed and supported here. For other functions, some extreme tests have been included.

zhx828: You don't know where it's going to call this function, so the input is not always guaranteed. If we have typescript, these do not need to be tested. But for now, let's have some test coverages.

web/yo/app/scripts/services/drugmaputils.js
```diff
@@ -11,9 +11,11 @@ angular.module('oncokbApp')
     .factory('drugMapUtils', function(OncoKB, _, $q, DatabaseConnector, $rootScope, ReviewResource, S, UUIDjs, $routeParams, firebaseConnector, firebasePathUtils) {
 
         function therapyStrToArr(key){
-            return key.split(",").map(function(element){
-                return element.trim().split(" + ")
-            });
+            if(key){
```
zhx828: This will return undefined for condition not passed, is this expected?

web/yo/app/scripts/services/mainutils.js
```diff
@@ -285,18 +285,29 @@ angular.module('oncokbApp')
...
-                param,
-                function(result) {
-                    deferred.resolve(result);
-                },
-                function(result) {
-                    deferred.reject(result);
-                }
-            );
-            return deferred.promise;
+            if(content){
```
zhx828: Also do a sendTo check.

web/yo/app/scripts/services/drugmaputils.js
Yizhuzhu: Yes. Generally, it should always be passed, because the key is checked before calling this function. I set this condition just for testing. If not, split will lead to error.

web/yo/app/scripts/services/mainutils.js
Yizhuzhu: I think check might not be needed here, because there is a promise. If sendTo is null, the return should be reject. If I did check here, there might not be reminder.

zhx828: You should handle the empty/undefined case and reject the promise manually. The promise you referenced is the one actually sending to backend for the service. But the whole purpose of checking is to avoid that.

web/yo/app/scripts/services/drugmaputils.js
zhx828: Should it return undefined or return empty string.

Issue #1619, opened 2019-06-03
Incorrect annotation for the query


zhx828:
The query should not get any annotations 
```
{
  "query": {
    "id": null,
    "type": null,
    "hugoSymbol": "KIT",
    "entrezGeneId": 3815,
    "alteration": "",
    "alterationType": null,
    "svType": null,
    "tumorType": "breast invasive ductal carcinoma",
    "consequence": "missense_variant",
    "proteinStart": null,
    "proteinEnd": null,
    "hgvs": null
  }
}
```

PR #1622, opened 2019-06-04
Fix issue when alteration is empty but consequence is specified

zhx828:
This solves https://github.com/oncokb/oncokb/issues/1619

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -658,7 +658,7 @@ public final class AlterationUtils {
         }
 
         if (alteration.getConsequence() != null && alteration.getConsequence().equals(VariantConsequenceUtils.findVariantConsequenceByTerm("missense_variant"))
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -657,7 +657,8 @@ public static String getInferredAlterationsKnownEffect(String inferredAlt) {
             return new ArrayList<>();
         }
 
-        if (alteration.getConsequence() != null && alteration.getConsequence().equals(VariantConsequenceUtils.findVariantConsequenceByTerm("missense_variant"))) {
+        if (alteration.getConsequence() != null && alteration.getConsequence().equals(VariantConsequenceUtils.findVariantConsequenceByTerm("missense_variant"))
+            && !alteration.getProteinStart().equals(-1) && !alteration.getProteinStart().equals(100000)) {
```
victoria34: What's the meaning of -1 and 100000 in here?

zhx828: It's the default of protein start/end if it does not have any information. This probably should be set to a constant. Let me update the PR.

PR #1661, opened 2019-07-18
Downward looking for diagnostic summmary/implication but upward looin…

zhx828:
…g for prognostic

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
@@ -198,7 +198,8 @@ public class IndicatorUtils {
 
             List<Alteration> nonVUSRelevantAlts = AlterationUtils.excludeVUS(relevantAlterations);
             Map<String, LevelOfEvidence> highestLevels = new HashMap<>();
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -198,7 +198,8 @@ public static IndicatorQueryResp processQuery(Query query, String geneStatus,
 
             List<Alteration> nonVUSRelevantAlts = AlterationUtils.excludeVUS(relevantAlterations);
             Map<String, LevelOfEvidence> highestLevels = new HashMap<>();
-            List<TumorType> oncoTreeTypes = new ArrayList<>();
+            List<TumorType> relevantUpwardTumorTypes = new ArrayList<>();
```
victoria34: Could you give a few comments to explain the `relevantUpwardTumorTypes` and `relevantDownwardTumorTypes`? Thanks

zhx828: upward/downward is the relevant concept to oncotree. Getting all parent nodes with same main type in oncotree is upward. Getting all child nodes with same main type in oncotree is downward.

PR #1664, opened 2019-07-23
Revert "Show uuid in api returned data"

victoria34:
Reverts oncokb/oncokb#1662


victoria34:
The code is already reverted so this PR is useless.

Issue #1779, opened 2019-10-29
SEVERE: Unable to process Jar entry [module-info.class] from Jar


zhx828:
SEVERE: Unable to process Jar entry [module-info.class] from Jar [jar:file:/target/tomcat.8888/webapps/expanded/WEB-INF/lib/log4j-api-2.10.0.jar!/] for annotations
org.apache.tomcat.util.bcel.classfile.ClassFormatException: Invalid byte tag in constant pool: 19
	at org.apache.tomcat.util.bcel.classfile.Constant.readConstant(Constant.java:97)
	at org.apache.tomcat.util.bcel.classfile.ConstantPool.<init>(ConstantPool.java:55)
	at org.apache.tomcat.util.bcel.classfile.ClassParser.readConstantPool(ClassParser.java:176)
	at org.apache.tomcat.util.bcel.classfile.ClassParser.parse(ClassParser.java:85)
	at org.apache.catalina.startup.ContextConfig.processAnnotationsStream(ContextConfig.java:2011)
	at org.apache.catalina.startup.ContextConfig.processAnnotationsJar(ContextConfig.java:1961)
	at org.apache.catalina.startup.ContextConfig.processAnnotationsUrl(ContextConfig.java:1936)
	at org.apache.catalina.startup.ContextConfig.processAnnotations(ContextConfig.java:1897)
	at org.apache.catalina.startup.ContextConfig.webConfig(ContextConfig.java:1149)
	at org.apache.catalina.startup.ContextConfig.configureStart(ContextConfig.java:771)
	at org.apache.catalina.startup.ContextConfig.lifecycleEvent(ContextConfig.java:305)
	at org.apache.catalina.util.LifecycleSupport.fireLifecycleEvent(LifecycleSupport.java:95)
	at org.apache.catalina.util.LifecycleBase.fireLifecycleEvent(LifecycleBase.java:90)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5154)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1408)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1398)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #1781, opened 2019-10-30
fix a compiling issue Unable to process Jar entry

zhx828:
Solution mentionedin https://github.com/jboss-javassist/javassist/issues/29

Signed-off-by: Hongxin Zhang <hongxin@cbio.mskcc.org>


zhx828:
This solves https://github.com/oncokb/oncokb/issues/1779

Issue #1799, opened 2019-10-29
Incorrect mapping


ritikakundra:
Wrong genomic criteria are matching. The fusion is matching to incorrect genomic criteria:
<img width="428" alt="Screen Shot 2019-10-29 at 4 25 19 PM" src="https://user-images.githubusercontent.com/12007862/67806269-ebbbb180-fa68-11e9-8440-b1086a451029.png">

victoria34:
Hi @zhx828, could you also take a look of this issue? I tried that in Postman. The API does match those alterations.

![image](https://user-images.githubusercontent.com/14971266/68143700-01fdbd80-ff00-11e9-96b0-23f2f263d174.png)


API : ....../match/variant
Input: 
```
{
  "oncokbVariants": [
    {
      "alteration": "Fusions",
      "hugoSymbol": "BRAF"
    },
    {
      "alteration": "T599TT",
      "hugoSymbol": "BRAF"
    },
    {
      "alteration": "N486_T491delinsK",
      "hugoSymbol": "BRAF"
    },
    {
      "alteration": "K601T",
      "hugoSymbol": "BRAF"
    },
    {
      "alteration": "T599TS",
      "hugoSymbol": "BRAF"
    },
    {
      "alteration": "Kinase Domain Duplication",
      "hugoSymbol": "BRAF"
    }
  ],
  "queries": [
    {
      "alteration": "BRAF-MKRN1 fusion",
      "hugoSymbol": "BRAF",
      "id": "P-0016986-T01-IM6"
    }
  ]
}
```

PR #1803, opened 2019-11-04
getRevertFusions should use provided fullAlterations when needed

zhx828:
This solves https://github.com/oncokb/oncokb/issues/1799

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -289,8 +289,27 @@ public final class AlterationUtils {
         return flag;
     }
 
+    public static Alteration getRevertFusions(Alteration alteration, Set<Alteration> fullAlterations) {
+        if (fullAlterations == null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -289,8 +289,27 @@ public static Boolean isFusion(String variant) {
         return flag;
     }
 
+    public static Alteration getRevertFusions(Alteration alteration, Set<Alteration> fullAlterations) {
+        if (fullAlterations == null) {
+            return getRevertFusions(alteration);
+        } else {
+            String revertFusionAltStr = getRevertFusionName(alteration);
```
victoria34: Could you explain the logic in here a little bit? What does match.isPresent() mean?

zhx828: @victoria34 this is a new method in Java 8, https://www.baeldung.com/java-optional. In here, I do a filter through stream, it produces an optional. If there is a match, then the isPresent is true.

Issue #1802, opened 2019-11-04
The annotation is not expected


zhx828:
The link below should return likely oncogenic

https://beta.oncokb.org/api/v1/annotate/mutations/byProteinChange?hugoSymbol=PIK3R1&alteration=W583del&tumorType=UCS&consequence=splice_region_variant

PR #1804, opened 2019-11-04
When get exact matched alteration, check consequence

zhx828:
Fix https://github.com/oncokb/oncokb/issues/1802 introduced in 1.5


victoria34:
Could you add the ticket link if it exists?

zhx828:
@victoria34 there is no ticket. I found the issue while testing oncokb-annotator.

zhx828:
@victoria34 actually I created earlier. Updated the comment

PR #1820, opened 2019-11-11
Upgrade oncotree to 2019_08_01

zhx828:


victoria34:
We need to upgrade oncotree of trial curation platform as well.

zhx828:
@victoria34 could we use oncotree from oncokb?

Issue #1793, opened 2019-10-29
Amp, point mutations, fusions etc should fall under oncogenic mutations


ritikakundra:
All alterations that have been tagged as oncogenic in oncokb should be matched for trials with oncogenic mutations genomic criteria

PR #1862, opened 2019-12-16
Fix amplification cannot matched to oncogenic mutations

victoria34:
Fix https://github.com/oncokb/oncokb/issues/1793


zhx828:
@victoria34 we are not going to map Oncogenic Mutations to Amplification. The trial curation will include Oncogenic Mutations and Amplification together.

PR #1914, opened 2020-02-14
Update frontend.version

victoria34:


victoria34:
> I always use 10 digits, would 7 work?

Yes, the 7 digits code works at this time but I have updated to 10 digits in case it may have the same shortcode with other commits.

PR #2353, opened 2020-12-07
update search/variants/clinical to return dx/px envidences

Leowisd:
This related to https://github.com/oncokb/oncokb/issues/2352


core/src/main/java/org/mskcc/cbio/oncokb/util/EvidenceTypeUtils.java
```diff
@@ -59,7 +59,8 @@
         types.add(EvidenceType.STANDARD_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_RESISTANCE);
         types.add(EvidenceType.INVESTIGATIONAL_THERAPEUTIC_IMPLICATIONS_DRUG_RESISTANCE);
         types.add(EvidenceType.INVESTIGATIONAL_THERAPEUTIC_IMPLICATIONS_DRUG_SENSITIVITY);
-
+        types.add(EvidenceType.DIAGNOSTIC_IMPLICATION);
+        types.add(EvidenceType.PROGNOSTIC_IMPLICATION);
```
zhx828, line 63: These two are note treatment evidences. Might just need to add these two at the place calling this util. Or create a util to return all. Maybe call getImplicationEvidenceTypes

zhx828 (changes requested):

zhx828 (approved):

PR #2727, opened 2021-10-11
Support nonstop/stop lost mutation format

### LINKED ISSUE #2724: Support nonstop mutation protein change
The format of the nonstop mutation needs to be supported.
For instance, DIS3 `*959Qext*14`

zhx828:
We are following the format described in https://varnomen.hgvs.org/recommendations/protein/variant/extension/
This fixes #2724


zhx828:
> Looks good. It could also be as simple as starting with "*"

Oh interesting. I will keep an eye on it if other formats arise.

Issue #2772, opened 2021-10-28
Dynamic redis cache squeezes out other cache keys when reach to the threshold


zhx828:
The redis size threshold is 3GB, we noticed that the `oncokb-core:processQuery` keeps pushing out other caches since it's mostly accessed and also keeps increasing.

Issue #2659, opened 2021-08-03
Update the eviction policy in Redisson


zhx828:
The eviction policy in Redis only applies to the key/value, but not hashmap. In production, OncoKB uses one Redis for all instances caching, we had to create hashmaps for different instance for better management.

In order to evict the cache, we now should use the MapCache(https://github.com/redisson/redisson/wiki/7.-distributed-collections#711-map-eviction-local-cache-and-data-partitioning) instead of the Map in Redisson.

- [ ] Set the # of keys to 2M for the query-indicator collection.
- [ ] Set the # of keys to 1M for the GN collection.
- [ ] LFU seems a better policy for OncoKB

zhx828:
We now in production use bucket cache, so all keys are on the same level before having a comprehensive solution.

PR #2782, opened 2021-11-04
Use redisson bucket cache for all caches

zhx828:
The map cache is easy to structure the relevant caches together but needs better eviction policy to get rid of the caches when reach threshold. Before having that, we should switch to bucket cache across all categories.

This fixes https://github.com/oncokb/oncokb/issues/2772 and is related to https://github.com/oncokb/oncokb/issues/2659


calvinlu3 (approved):
Noted the changes and everything looks fine. 👍

PR #2827, opened 2021-12-17
Map additional splicing variant

zhx828:
Map all splicing variant listed under http://sequenceontology.org/browser/current_svn/term/SO:0001568 to splice_region_variant for better annotation


zhx828:
The main use case for this pr is to support additional consequences from VEP. But since we started to use GN variant class for unmapped consequence(https://github.com/oncokb/oncokb/pull/2841), this pr is no longer necessary.

PR #2835, opened 2021-12-21
Fix the issue when both mostSevereVariantConsequence and consequenceTerms cannot be mapped

zhx828:


calvinlu3 (approved):
Just tested the following locally:
- `/api/v1/annotate/mutations/byHGVSg?hgvsg=5%3Ag.67589635_67589639delinsTA&referenceGenome=GRCh37`
- `/api/v1/annotate/mutations/byHGVSg?hgvsg=4%3Ag.55593600_55593606delinsGTGG&referenceGenome=GRCh37`

Getting valid responses. 👍

PR #2840, opened 2021-12-29
Do not use mostSevereConsequence to choose the transcript

zhx828:
The mostSevereConsequence maybe different from the matched transcript. Example: https://grch38.genomenexus.org/variant/7:g.140753336A%3ET The most server is stop_lost but it really should be missense_variant

This fixes https://github.com/oncokb/oncokb-annotator/issues/148

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
@@ -158,17 +158,16 @@ public class GenomeNexusUtils {
         return variantAnnotation;
     }
 
-    public static VariantConsequence getTranscriptConsequenceSummaryTerm(String consequenceTerms, String mostSevereConsequence) {
+    public static VariantConsequence getTranscriptConsequenceSummaryTerm(String consequenceTerms) {
         if (StringUtils.isEmpty(consequenceTerms)) {
             return null;
         }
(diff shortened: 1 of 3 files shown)


ZWael:
Hi @zhx828,
in the provided link (BRAF V600E mutation) > All transcripts table > Consequence Terms i see the missense_variant but no stop_lost 
can you explain your reasoning

zhx828:
@ZWael sorry, put the wrong example, just updated the comment. 
We use the annotation endpoint from genome-nexus, if you look at the response, https://www.genomenexus.org/annotation/3%3Ag.38182641T%3EC, the `most_severe_consequence` is `stop_lost`. But rather, we should use the consequence_terms of transcript `ENST00000396334`

ZWael:
@zhx828  thank you for the feedback and the explanations, 
I see that the Variant Classification is a Missense_Mutation, Consequence Terms : missense_variant
but does it explain the oncokb annotations : "MUTATION_EFFECT : Unknown/Gain-of-function"  "ONCOGENIC:  Predicted Oncogenic/Oncogenic" depending of the query type ?

thanks for the genome nexus resource

zhx828:
> 

Yup yup, it does. The fix is in production now.

zhx828:
> @zhx828 thank you for the feedback and the explanations, I see that the Variant Classification is a Missense_Mutation, Consequence Terms : missense_variant but does it explain the oncokb annotations : "MUTATION_EFFECT : Unknown/Gain-of-function" "ONCOGENIC: Predicted Oncogenic/Oncogenic" depending of the query type ?
> 
> thanks for the genome nexus resource

Yup yup, it does. The fix is in production now.

Issue #2897, opened 2022-01-18
Add software version in the api info endpoint


zhx828:
Since the software changes the annotation sometimes, we should expose such info in the endpoint. Ideally, it should be the oncokb-core released version.

PR #2948, opened 2022-02-09
Include appVersion in the info endpoint

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/2897

The data looks like the following. I'm using https://github.com/vdurmont/semver4j
```
appVersion: {
  originalValue: "3.6.3",
  value: "3.6.3",
  major: 3,
  minor: 6,
  patch: 3,
  suffixTokens: [ ],
  build: null,
  type: "STRICT",
  stable: true
}
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -175,19 +175,13 @@
         <plugins>
           <!-- any other plugins -->
           <plugin>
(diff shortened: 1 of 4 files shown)


zhx828:
@calvinlu3 updated the comment above with the data structure.

Issue #3192, opened 2022-07-13
Update Swagger v2 host


zhx828:
The base URL in the current swagger shows `oncokb-core:8888/api/v1` https://www.oncokb.org/swagger-ui/index.html#/
But it should be www.oncokb.org

PR #3264, opened 2022-09-19
Add property to update swagger hostname

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3192

- Added a new property `swagger_hostname`
- When property is not set, then the default host will be supplied by the swagger config. When property is set, then the default will be overriden.

image

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/config/MvcConfigurationPublic.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/config/MvcConfigurationPublic.java
@@ -1,7 +1,8 @@
 package org.mskcc.cbio.oncokb.config;
 
-import com.mysql.jdbc.StringUtils;
 import io.sentry.spring.SentryExceptionResolver;
+
+import org.apache.commons.lang3.StringUtils;
 import org.mskcc.cbio.oncokb.config.annotation.PremiumPublicApi;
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/config/MvcConfigurationPublic.java
```diff
@@ -68,6 +71,22 @@ public Docket publicApi() {
                 "https://www.oncokb.org/terms"
             ))
             .useDefaultResponseMessages(false);
+        
+        // Update swagger host with environment variable
+        String swaggerHost = PropertiesUtils.getProperties("swagger_hostname");
+        if (!StringUtils.isNullOrEmpty(swaggerHost)) {
```
zhx828, line 77: I remember it has a method isNotEmpty?

zhx828 (commented):
This PR needs to go to next-minor-release branch. It will trigger api client changes.

web/src/main/java/org/mskcc/cbio/oncokb/config/MvcConfigurationPublic.java
calvinlu3: The file already had StringUtils imported from a different library: `com.mysql.jdbc`. I changed to use the StringUtils from apache library which has the isNotEmpty method.

PR #3281, opened 2022-10-20
Add redis connection size properties and upgrade Redisson to v3.17.7

calvinlu3:
- This will make it easier to change the `slaveConnectionMinimumIdleSize`, `slaveConnectionPoolSize`, `masterConnectionMinimumIdleSize`, `masterConnectionPoolSize` Redis properties.

- Upgrade Redisson from 3.15.5 to 3.17.7
  - There are some fixes related to Redis Cluster
- Add client name

### DIFF
--- a/core/src/main/resources/properties-EXAMPLE/config.properties
+++ b/core/src/main/resources/properties-EXAMPLE/config.properties
@@ -57,6 +57,11 @@ redis.password=oncokb-redis-password
 redis.masterName=oncokb-master
 # in minutes
 redis.expiration=30
+redis.slaveConnectionMinimumIdleSize=
+redis.slaveConnectionPoolSize=
+redis.masterConnectionMinimumIdleSize=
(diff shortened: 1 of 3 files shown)


core/src/main/resources/properties-EXAMPLE/config.properties
```diff
@@ -57,6 +57,11 @@ redis.password=oncokb-redis-password
 redis.masterName=oncokb-master
 # in minutes
 redis.expiration=30
+redis.slaveConnectionMinimumIdleSize=
+redis.slaveConnectionPoolSize=
+redis.masterConnectionMinimumIdleSize=
+redis.masterConnectionPoolSize=
+redis.clientName=
```
zhx828, line 64: Can we use `app.name` and add `-client` after?

zhx828 (changes requested):

zhx828 (approved):
👍🏼

PR #3280, opened 2022-10-19
Cache transcript genes

zhx828:
This is related to https://app.zenhub.com/workspaces/oncokb-568a9eeafc8895f63ab9e2f9/issues/oncokb/oncokb/3278

Issue #3276, opened 2022-10-18
oncokb-core performance improvement


zhx828:
This is an epic to keep tracking of performance improvements across multiple oncokb-core endpoints.

### Todo

- [x] Data update is functioning as expected (data propagation from one instance to another)

PR #3282, opened 2022-10-24
Overall performance improvements

zhx828:
This is an effort to improve the performance without caching after https://github.com/oncokb/oncokb/pull/3273 and https://github.com/oncokb/oncokb/pull/3280

This is related to https://github.com/oncokb/oncokb/issues/3276

## Test
The following result is generated based on one POST with 10 continues calls
Payload: 500 variants POST byProteinChange [payload.json.zip](https://github.com/oncokb/oncokb/files/9855684/payload.json.zip)

### Before 
Average per call: 6.4s
before

### After
Average per call: 2s
after

## Main changes
- Use alteration id as hash code when available
- Use tumor type id as hash code when available
- Use simple for loop (for-i) instead of for-each. This has significant improvement if we access items through index instead of using iterator. This is based on the base data is already unique in the collection.
- Use string regionMatches instead of regex https://github.com/oncokb/oncokb/pull/3282/files#diff-5cbe99e8fb6d33a48ccc0c3447037d0c9816dca3b995402fcfbfada9e0b0b70eR1207-R1243
- Cache evidence relevant cancer types
- Cache alteration also based on reference genome


core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -375,7 +367,22 @@ public static Boolean containAlterations(Integer entrezGeneId) {
 
     public static void setAlterations(Gene gene) {
         if (gene != null && genes.contains(gene)) {
-            alterations.put(gene.getEntrezGeneId(), new HashSet<>(ApplicationContextSingleton.getAlterationBo().findAlterationsByGene(Collections.singleton(gene))));
+            List<Alteration> geneAlterations = ApplicationContextSingleton.getAlterationBo().findAlterationsByGene(Collections.singleton(gene));
+            if (!alterations.containsKey(gene.getEntrezGeneId())) {
+                alterations.put(gene.getEntrezGeneId(), new ArrayList<>());
+            }
```
calvinlu3, line 373: Do we need to initialize the map with an empty ArrayList value if it is immediately replaced by `geneAlterations` at line 377?

calvinlu3 (approved):
Looks good

core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
zhx828: Makes sense. I will update.

PR #3336, opened 2022-12-21
Oncogenic Mutations should not be included as relevant alterations when alternative allele is resistance

zhx828:
When calculating relevant alterations, we should not associate Oncogenic Mutations when alternative allele is resistance.

This fixes https://github.com/oncokb/oncokb-pipeline/issues/75


core/src/main/java/org/mskcc/cbio/oncokb/model/IndicatorQueryOncogenicity.java
```diff
@@ -0,0 +1,19 @@
+package org.mskcc.cbio.oncokb.model;
+
+public class IndicatorQueryOncogenicity {
+    Oncogenicity oncogenicity;
```
calvinlu3, line 4: Since this got moved out of `IndicatorUtils.java` and into a different package, would it make sense to use the `private` access modifier instead of the default?

calvinlu3 (approved):
LGTM

core/src/main/java/org/mskcc/cbio/oncokb/model/IndicatorQueryOncogenicity.java
zhx828: Good point, let me fix

PR #3363, opened 2023-01-06
Remove send email controller and google api-client

zhx828:
The send email controller is not properly tested and does not have much usage. We should not allow core to send email.


core/src/main/resources/properties-EXAMPLE/config.properties
```diff
@@ -32,10 +32,6 @@ genome_nexus.grch38.url=https://grch38.genomenexus.org
 #Name of application
 app.name=
 
-#Google account info. They are used to send email (Optional)
-google.username=
-google.password=
-
 #where all log emails go
 curation.log.email.to=
```
calvinlu3, line 36: This can be removed too

calvinlu3 (approved):
In the README, we can remove the line
```
google.username & google.password(Optional) : Google account info. It is used to send email
```

Issue #3330, opened 2022-12-15
Verify whether HTTP 500 responses are related to redis cluster master node being down


calvinlu3:
#### Setup
- Redis Cluster with 3 master nodes and 3 replica nodes
- 1 oncokb-core-beta pod
- Use JMeter to send requests infinitely

#### Observations (when reading data that has already been cached)

When we make one of the master nodes unavailable (via pod deletion), the core is still able to function since the master node's data is replicated onto the replica and can still be read. The default `readMode` is `slave`, which tells Redisson to `Read from slave nodes, uses MASTER if no SLAVES are available.`

When both the master and its replica are unavailable, then we start getting HTTP 500 responses from the core. 
`org.redisson.client.WriteRedisConnectionException: Unable to write command into connection! Increase connection pool size. Node source: NodeSource [slot=6660, addr=null, redisClient=null, redirect=null, entry=null], connection: RedisConnection@1628860747 [redisClient=[addr=redis:&#47;&#47;10.1.202.42:6379], channel=[id: 0x3ebf3f4a, L:&#47;10.1.202.70:49318 ! R:10.1.202.42&#47;10.1.202.42:6379], currentCommand=null, usage=1], command: (GET), params: [oncokb-core-beta:general:processQuery:GRCh37:7157:null:R273C:null:Uterine Undifferentiated Carcinoma:Missense_Mutation:273:273:null:null:null:false:[]] after 3 retry attempts`
To avoid getting HTTP 500 responses when master+replica are down, we can possibly try to increase `retryAttempts` and `retryInterval` in Redisson config while Redis node attempts to restart?

<details><summary>Full API response</summary>

```
[...]

PR #3364, opened 2023-01-06
Allow fallback to non-cached version when redis cluster node is down

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3330

Changes:
- Add LoggingCacheErrorHandler to log the error messages when Redisson  GET, PUT, or CLEAR fails. This avoids our API returning 500 responses when  a redis cluster node may be down
- Allow redis lookup method to return `null` when Redis node is down (cache miss) to allow for fallback to non-cache-version

- [x] Do a test using beta

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/LoggingCacheErrorHandler.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/LoggingCacheErrorHandler.java
@@ -1,5 +1,7 @@
 package org.mskcc.cbio.oncokb.cache;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.cache.Cache;
 import org.springframework.cache.interceptor.CacheErrorHandler;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/cache/LoggingCacheErrorHandler.java
```diff
@@ -0,0 +1,39 @@
...
+ * that logs the error messages when performing Redis operations.
+ * Redis will throw a RuntimeException causing our APIs to return HTTP 500 responses, so we defined
+ * this class to just log the errors and allow our app fallback to the non-cached version.
+ */
+
+public class LoggingCacheErrorHandler implements CacheErrorHandler {
+
+    @Override
+    public void handleCacheGetError(RuntimeException exception, Cache cache, Object key) {
+        System.out.println(String.format("Cache '%s' failed to get entry with key '%s'", cache.getName(), key));
```
zhx828, line 17: Should we use sj4j logger to print?

zhx828 (approved):
Looks good!

Issue #3304, opened 2022-11-16
Import pre-annotated GN file to Redis


calvinlu3:
- [x] Create endpoint in oncokb-core to import the pre annotated text file contents to Redis cache
- [x] Python script to call endpoint to generate initial redis cache
- [x] Find size of data
   - For ~100K keys, the size of redis data was ~120MB (one node in 3 node cluster took 40MB, 40MB * 3 nodes = 120).
   - We have a total of ~2.2 million keys, which will take 2.6GB (2200000/100000 * 120 MB = 2640MB)
   - Since we have a 10 cluster node in production, each node will hold ~260MB of data
   - We cache two times (by hgvsg and gl), so 260MB * 2 = ~520MB

- [ ] See how we can dump redis cache when restarts and how to import back. (Automatically restore?)
### Option 1: Generate RDB files and restore
  1. After all variants are cached in Redis, loop through master nodes and call BGSAVE to create RDB file. (We should have 10 RDB files, 1 per shard)
  2. Use LASTSAVE redis command to determine if background save has finished
  3. We can store the RDB files to S3 (https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/connections/aws.html). Include timestamp 
  4.  To import the gn variants cache back, we just need to copy the RDB file into the redis directory (CONFIG GET dir).
  5. Restart redis server so that RDB file are read on startup. 
  
   https://docs.bitnami.com/kubernetes/infrastructure/redis/administration/backup-restore/

### Option 2: ✔ 
1. We have an airflow task to delete the caches when it is full (convert the cronjob to airflow)
2. We have another airflow task to call the API endpoint to cache the files from the annotated txt files.
  - Have two airflow DAGs: low priority(every 30 min) and high priority (every 5 mins), cutoff is sample_count of 10

- [x] Find distribution of variants files w/ sample count
```
Total: 2669255
                  counts
sample_count
(0, 1]            613471
(1, 2]            580746
(2, 3]            360242
(3, 4]            295359
(4, 5]            154608
(5, 10]           448129
(10, 100]         211580
(100, 500]          4326
(500, 1000]          436
(1000, 5000]         293
(5000, 10000]         34
(10000, 50000]        28
(50000, 100000]        2
(100000, 500000]       1
```

PR #3380, opened 2023-01-25
Add endpoint to cache pre-annotated gn variants

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3304
- Endpoint to import annotated variants into redis cache: https://github.com/oncokb/oncokb/pull/3380/commits/d1c5af84092938d5d4f4959d5a5ba8cae939c82e
- GN annotation API provides the `hgsvg` and `genomicLocation` in response, so made changes to use those: https://github.com/oncokb/oncokb/pull/3380/commits/544da19cc50c38bdcb4b5d952c651239ec54d954
  - We can't use [GenomicLocation.toString()](https://github.com/genome-nexus/genome-nexus/blob/master/model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java#L63-L65) because the swagger code generator override the [toString()](https://github.com/genome-nexus/genome-nexus-java-api-client/blob/master/genomeNexusPublicApiClient/src/main/java/org/genome_nexus/client/GenomicLocation.java#L159-L171). The gl string was built manually from object.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
@@ -253,8 +253,8 @@ public class CacheFetcher {
         Alteration alteration = new Alteration();
 
         // Build the Alteration object from the pre-annotated GN variant object.
-        if (StringUtils.isNotEmpty(gnAnnotatedVariantInfo.getHugoSymbol())) {
-            Gene gene = GeneUtils.getGene(gnAnnotatedVariantInfo.getHugoSymbol());
+        if (StringUtils.isNotEmpty(gnAnnotatedVariantInfo.getHugoSymbol()) && gnAnnotatedVariantInfo.getEntrezGeneId() != null) {
+            Gene gene = GeneUtils.getGene(gnAnnotatedVariantInfo.getEntrezGeneId(), gnAnnotatedVariantInfo.getHugoSymbol());
             if (gene == null) {
                 gene = new Gene();
                 gene.setHugoSymbol(gnAnnotatedVariantInfo.getHugoSymbol());
@@ -274,12 +274,12 @@ public class CacheFetcher {
         // Store pre-annotated alteration into Redis cache
         Cache cache = cacheManager.getCache(CacheCategory.GENERAL.getKey() + REDIS_KEY_SEPARATOR + "getAlterationFromGenomeNexus");
         if (StringUtils.isNotEmpty(hgvsg)) {
-            String cacheKey = org.springframework.util.StringUtils.arrayToDelimitedString(new Object[]{GNVariantAnnotationType.HGVS_G, referenceGenome, hgvsg}, REDIS_KEY_SEPARATOR);
+            String cacheKey = String.join(REDIS_KEY_SEPARATOR, new String[]{GNVariantAnnotationType.HGVS_G.name(), referenceGenome.name(), hgvsg});
             cache.putIfAbsent(cacheKey, alteration);
         }
 
         if (StringUtils.isNotEmpty(genomicLocation)) {
-            String cacheKey = org.springframework.util.StringUtils.arrayToDelimitedString(new Object[]{GNVariantAnnotationType.GENOMIC_LOCATION, referenceGenome, genomicLocation}, REDIS_KEY_SEPARATOR);
+            String cacheKey = String.join(REDIS_KEY_SEPARATOR, new String[]{GNVariantAnnotationType.GENOMIC_LOCATION.name(), referenceGenome.name(), genomicLocation});
             cache.putIfAbsent(cacheKey, alteration);
         }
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
```diff
@@ -159,11 +159,18 @@ public static GenomeNexusAnnotatedVariantInfo getAnnotatedVariantFromGenomeNexus
...
+            if (StringUtils.isNotEmpty(annotation.getHgvsg())) {
+                preAnnotatedVariantInfo.setHgvsg(annotation.getHgvsg());
+            }
 
-            if (type.equals(GNVariantAnnotationType.GENOMIC_LOCATION) && annotation.isSuccessfullyAnnotated()) {
-                preAnnotatedVariantInfo.setGenomicLocation(query);
+            if (annotation.getAnnotationSummary() != null && annotation.getAnnotationSummary().getGenomicLocation() != null) {
+                org.genome_nexus.client.GenomicLocation gl = annotation.getAnnotationSummary().getGenomicLocation();
+                String glString = gl.getChromosome() + "," + gl.getStart() + "," + gl.getEnd() + "," + gl.getReferenceAllele() + "," + gl.getVariantAllele();
+                if (StringUtils.isNotEmpty(glString)) {
```
zhx828, line 171: glString will never be empty.

zhx828: We should validate on all five data elements.

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -244,6 +249,42 @@ public Alteration getAlterationFromGenomeNexus(GNVariantAnnotationType gnVariant
         return AlterationUtils.getAlterationFromGenomeNexus(gnVariantAnnotationType, genomicLocation, referenceGenome);
     }
 
+    public void cacheAlterationFromGenomeNexus(GenomeNexusAnnotatedVariantInfo gnAnnotatedVariantInfo) {
+        Alteration alteration = new Alteration();
+
+        // Build the Alteration object from the pre-annotated GN variant object.
+        if (StringUtils.isNotEmpty(gnAnnotatedVariantInfo.getHugoSymbol())) {
```
zhx828, line 256: Let's do check on either hugo symbol or entrez gene id

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -244,6 +249,42 @@ public Alteration getAlterationFromGenomeNexus(GNVariantAnnotationType gnVariant
...
+            Gene gene = GeneUtils.getGene(gnAnnotatedVariantInfo.getHugoSymbol());
```
zhx828, line 257: we should find using entrez gene id first, then hugo

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -244,6 +249,42 @@ public Alteration getAlterationFromGenomeNexus(GNVariantAnnotationType gnVariant
...
+        alteration.setProteinStart(gnAnnotatedVariantInfo.getProteinStart());
+        alteration.setProteinEnd(gnAnnotatedVariantInfo.getProteinEnd());
+        alteration.setConsequence(VariantConsequenceUtils.findVariantConsequenceByTerm(gnAnnotatedVariantInfo.getConsequenceTerms()));
+
+        String hgvsg = gnAnnotatedVariantInfo.getHgvsg();
+        String genomicLocation = gnAnnotatedVariantInfo.getGenomicLocation();
+        ReferenceGenome referenceGenome = gnAnnotatedVariantInfo.getReferenceGenome();
+
+        // Store pre-annotated alteration into Redis cache
+        Cache cache = cacheManager.getCache(CacheCategory.GENERAL.getKey() + REDIS_KEY_SEPARATOR + "getAlterationFromGenomeNexus");
```
zhx828, line 275: Can we use ConcatGenerator?

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -244,6 +249,42 @@ public Alteration getAlterationFromGenomeNexus(GNVariantAnnotationType gnVariant
...
+        if (StringUtils.isNotEmpty(hgvsg)) {
+            String cacheKey = org.springframework.util.StringUtils.arrayToDelimitedString(new Object[]{GNVariantAnnotationType.HGVS_G, referenceGenome, hgvsg}, REDIS_KEY_SEPARATOR);
```
zhx828, line 277: I would specify the content being String. `new String[]{A, B}`, then do a String.join();

zhx828 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
calvinlu3: I don't think we can because ConcatGenerator only generates the part of the key with the method params. 
```
public class ConcatGenerator implements KeyGenerator {

    public Object generate(Object target, Method method, Object... params) {
        return StringUtils.arrayToDelimitedString(params, REDIS_KEY_SEPARATOR);
    }
}
```
I am fetching the cache from the cache manager with the hardcoded name `CacheCategory.GENERAL.getKey() + REDIS_KEY_SEPARATOR + "getAlterationFromGenomeNexus"`. Not sure if there's a better way to figure out which cache we should put the values into.

zhx828 (approved):

PR #3388, opened 2023-02-02
Modify trial data format and trials endpoint for compatibility b/w AACT and OncoKB

scarrero4660:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/97
Updated data is also added in this [issue](https://github.com/oncokb/oncokb-pipeline/issues/97)

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
@@ -5,6 +5,8 @@ import org.cbioportal.genome_nexus.component.annotation.NotationConverter;
 import org.cbioportal.genome_nexus.model.GenomicLocation;
 import org.cbioportal.genome_nexus.util.exception.InvalidHgvsException;
 import org.cbioportal.genome_nexus.util.exception.TypeNotSupportedException;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
 import org.mskcc.cbio.oncokb.apiModels.CuratedGene;
 import org.mskcc.cbio.oncokb.bo.OncokbTranscriptService;
 import org.mskcc.cbio.oncokb.genomenexus.GNVariantAnnotationType;
@@ -21,8 +23,17 @@ import org.springframework.cache.annotation.Cacheable;
 import org.springframework.stereotype.Component;
 
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.util.*;
 import java.util.stream.Collectors;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.services.s3.model.S3Object;
+import com.amazonaws.services.s3.model.S3ObjectInputStream;
 
 import static org.mskcc.cbio.oncokb.Constants.DEFAULT_REFERENCE_GENOME;
 import static org.mskcc.cbio.oncokb.util.MainUtils.rangesIntersect;
@@ -32,6 +43,15 @@ import static org.mskcc.cbio.oncokb.cache.Constants.REDIS_KEY_SEPARATOR;
 public class CacheFetcher {
     OncokbTranscriptService oncokbTranscriptService = new OncokbTranscriptService();
     NotationConverter notationConverter = new NotationConverter();
+    
+    final String s3AccessKey = PropertiesUtils.getProperties("aws.s3.accessKey");
+    final String s3SecretKey = PropertiesUtils.getProperties("aws.s3.secretKey");
+    final String s3Region = PropertiesUtils.getProperties("aws.s3.region");
+    AWSCredentials credentials = new BasicAWSCredentials(s3AccessKey, s3SecretKey);
+            AmazonS3 s3client = AmazonS3ClientBuilder.standard()
+                .withCredentials(new AWSStaticCredentialsProvider(credentials))
+                .withRegion(s3Region)
+                .build();
 
     @Autowired(required = false) 
     CacheManager cacheManager;
@@ -342,5 +362,31 @@ public class CacheFetcher {
             }
         }).collect(Collectors.toList());
         return filtered.size() > 0;
+    }
+
+    @Cacheable(cacheResolver = "generalCacheResolver")
+    public JSONObject getTrialsJSON() throws Exception {
+        try {
+            S3Object s3objectTrials = s3client.getObject("oncokb", "clinical-trials/results/trials.json");
+            S3ObjectInputStream inputStreamTrials = s3objectTrials.getObjectContent();
+            JSONParser jsonParser = new JSONParser();
+            return (JSONObject) jsonParser.parse(new InputStreamReader(inputStreamTrials, "UTF-8"));
+        } catch(Exception e) {
+            System.out.print("There was an error fetching the trials.");
+            throw new Exception(e);
+        }
+    }
+
+    @Cacheable(cacheResolver = "generalCacheResolver")
+    public JSONObject getOncotreeJSON() throws Exception {
+        try {
+            S3Object s3objectOncotree = s3client.getObject("oncokb", "clinical-trials/results/oncotree_mapping.json");
+            S3ObjectInputStream inputStreamOncotree = s3objectOncotree.getObjectContent();
+            JSONParser jsonParser = new JSONParser();
+            return (JSONObject) jsonParser.parse(new InputStreamReader(inputStreamOncotree, "UTF-8"));
+        } catch(Exception e) {
+            System.out.print("There was an error fetching the oncotree mapping.");
+            throw new Exception(e);
+        }
     }
 }
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
@@ -24,7 +24,6 @@ import java.io.IOException;
 import java.util.*;
 import java.util.stream.Collectors;
 
-
 import static org.mskcc.cbio.oncokb.Constants.DEFAULT_REFERENCE_GENOME;
 import static org.mskcc.cbio.oncokb.util.MainUtils.rangesIntersect;
 import static org.mskcc.cbio.oncokb.cache.Constants.REDIS_KEY_SEPARATOR;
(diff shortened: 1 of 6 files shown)


web/pom.xml
```diff
@@ -290,7 +290,7 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk</artifactId>
-            <version>1.12.377</version>
```
zhx828, line 293: why we downgrading the version?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
```diff
@@ -64,14 +66,19 @@ public ResponseEntity<List<Trial>> trialsMatchingGet(
...
 
-        S3Object s3object = s3client.getObject("oncokb", "drug-matching/result.json");
-        S3ObjectInputStream inputStream = s3object.getObjectContent();
+        S3Object s3objectTrials = s3client.getObject("oncokb-clinical-trials", "precision_med_trial_data.json");
+        S3ObjectInputStream inputStreamTrials = s3objectTrials.getObjectContent();
         JSONParser jsonParser = new JSONParser();
-        JSONObject jsonObject = (JSONObject) jsonParser.parse(new InputStreamReader(inputStream, "UTF-8"));
+        JSONObject jsonObjectTrials = (JSONObject) jsonParser.parse(new InputStreamReader(inputStreamTrials, "UTF-8"));
+        
+        S3Object s3objectOncotree = s3client.getObject("oncokb-clinical-trials", "oncotree_to_trials_precision_med.json");
```
zhx828, line 74: The naming of these files should be standardized. `precision_med` should be prefix or postfix, but not the mix usage.

zhx828: also the name `oncotree_to_trials` does not make sense

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
```diff
@@ -151,12 +163,31 @@ public ResponseEntity<Map<String, List<Trial>>> trialsGetByCancerTypes(
...
         Tumor tumor = new Tumor();
-        if (jsonObject.containsKey(oncoTreeCode)) {
-            JSONObject tumorObj = (JSONObject) jsonObject.get(oncoTreeCode);
+        if (jsonObjectOncotree.containsKey(oncoTreeCode)) {
+            JSONObject tumorObj = (JSONObject) jsonObjectOncotree.get(oncoTreeCode);
             Gson gson = new Gson();
+            // System.out.println(tumorObj.toString());
             tumor = gson.fromJson(tumorObj.toString(), Tumor.class);
+
+            List<Object> trialIds = (List<Object>) tumorObj.get("trials");
```
zhx828, line 174: This is not only ids? The naming should be updated.

zhx828 (changes requested):

web/pom.xml
scarrero4660: There was an issue previously where getting the data from the bucket would not work with the version we had if you recall

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
scarrero4660: Yes that was mainly for my testing purposes. I will update the names

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
scarrero4660: At this point the trial objects do not have any other information besides IDs but yes the naming does not make sense. I will update this

web/pom.xml
scarrero4660: Actually I tested it out with 1.12 and cannot recreate the error so I will update it back to 1.12

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -297,4 +317,30 @@ public boolean genomicLocationShouldBeAnnotated(GNVariantAnnotationType gnVarian
...
+            JSONParser jsonParser = new JSONParser();
+            return (JSONObject) jsonParser.parse(new InputStreamReader(inputStreamTrials, "UTF-8"));
+        } catch(Exception e) {
+            System.out.print("There was an error fetching the trials.");
+            throw new Exception(e);
+        }
+    }
+
+    @Cacheable(cacheResolver = "generalCacheResolver")
+    public JSONObject getOncotreeJSON() throws Exception {
```
zhx828, line 335: We need to rename this. Maybe `getOncoTreeMappingTrials`

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -19,16 +21,34 @@
...
     NotationConverter notationConverter = new NotationConverter();
+    
+    final String s3AccessKey = PropertiesUtils.getProperties("aws.s3.accessKey");
+    final String s3SecretKey = PropertiesUtils.getProperties("aws.s3.secretKey");
+    final String s3Region = PropertiesUtils.getProperties("aws.s3.region");
+    AWSCredentials credentials = new BasicAWSCredentials(s3AccessKey, s3SecretKey);
+            AmazonS3 s3client = AmazonS3ClientBuilder.standard()
+                .withCredentials(new AWSStaticCredentialsProvider(credentials))
+                .withRegion(s3Region)
+                .build();
```
zhx828, line 51: Can we move this to an util so we can reuse in other classes?

core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -222,6 +240,24 @@ private static void notifyOtherServices(String cmd, Set<Integer> entrezGeneIds)
...
+            final String s3AccessKey = PropertiesUtils.getProperties("aws.s3.accessKey");
+            final String s3SecretKey = PropertiesUtils.getProperties("aws.s3.secretKey");
+            final String s3Region = PropertiesUtils.getProperties("aws.s3.region");
+
+            AWSCredentials credentials = new BasicAWSCredentials(s3AccessKey, s3SecretKey);
+            AmazonS3 s3client = AmazonS3ClientBuilder.standard()
+                .withCredentials(new AWSStaticCredentialsProvider(credentials))
+                .withRegion(s3Region)
+                .build();
+            S3Object s3objectTrials = s3client.getObject("oncokb", "clinical-trials/results/trials.json");
```
zhx828, line 252: These file locations should be put on top as constants. The common folders should be shared to avoid typos.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
```diff
@@ -151,12 +151,30 @@ public ResponseEntity<Map<String, List<Trial>>> trialsGetByCancerTypes(
...
+    private Tumor getTumor(JSONObject jsonObjectOncotree, JSONObject jsonObjectTrials, String oncoTreeCode) {
         Tumor tumor = new Tumor();
-        if (jsonObject.containsKey(oncoTreeCode)) {
-            JSONObject tumorObj = (JSONObject) jsonObject.get(oncoTreeCode);
+        if (jsonObjectOncotree.containsKey(oncoTreeCode)) {
+            JSONObject tumorObj = (JSONObject) jsonObjectOncotree.get(oncoTreeCode);
             Gson gson = new Gson();
             tumor = gson.fromJson(tumorObj.toString(), Tumor.class);
+
+            List<Object> trials = (List<Object>) tumorObj.get("trials");
```
zhx828, line 161: What's the plan here if the obj does not have `trials`?

zhx828 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -53,6 +69,12 @@ public class CacheUtils {
     private static List<TumorType> mainTypes = new ArrayList<>();
     private static List<TumorType> specialCancerTypes = new ArrayList<>();
 
+    private static String oncokbS3Bucket = "oncokb";
+    private static String trialsS3Path = "clinical-trials/mappings/trials.json";
+    private static String oncotreeS3Path = "clinical-trials/mappings/oncotree_mapping.json";
```
zhx828, line 74: Constants should be all uppercase, for instance `ONCOKB_S3_BUCKET`.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
zhx828: Any comments on this?

core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -222,6 +244,25 @@ private static void notifyOtherServices(String cmd, Set<Integer> entrezGeneIds)
             System.out.println("Register other services: " + MainUtils.getTimestampDiff(current) + " at " + MainUtils.getCurrentTime());
             current = MainUtils.getCurrentTimestamp();
 
+            final String s3AccessKey = PropertiesUtils.getProperties("aws.s3.accessKey");
+            final String s3SecretKey = PropertiesUtils.getProperties("aws.s3.secretKey");
+            final String s3Region = PropertiesUtils.getProperties("aws.s3.region");
```
zhx828, line 249: What happens when these keys are not available? Will there be exceptions when initiating aws s3 client?

zhx828 (changes requested):

zhx828:
@scarrero4660 could you also rebase to the latest master and address the conflicts?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/TrialsApiController.java
scarrero4660: Currently if the oncotree code exists in the mapping, then there would not  be a case where the trials would be empty since the mapping is generated from the list of trials. If the query for the oncotree code is not available then an empty list would be returned

core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
scarrero4660: Yes I will add a check to handle this

core/src/main/java/org/mskcc/cbio/oncokb/util/CacheUtils.java
```diff
@@ -222,6 +241,17 @@ private static void notifyOtherServices(String cmd, Set<Integer> entrezGeneIds)
             System.out.println("Register other services: " + MainUtils.getTimestampDiff(current) + " at " + MainUtils.getCurrentTime());
             current = MainUtils.getCurrentTimestamp();
 
+            AmazonS3 s3client = MainUtils.startAWSClient();
```
zhx828, line 244:
```suggestion
            AmazonS3 s3client = MainUtils.getAWSClient();
```

zhx828 (approved):
LGTM, thanks!

PR #3411, opened 2023-03-01
Modify utils/filterHgvsgBasedOnCoverage endpoint to include filter status of all original queries

calvinlu3:
Instead of returning a list of strings of the filtered hgvsg/gl, we return an object that contains the hgvsg/gl and a boolean indicating whether or not it is filtered out.

This makes it easier in the gn pipeline script (https://github.com/oncokb/oncokb-pipeline/issues/107)

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/apiModels/TranscriptCoverageFilterResult.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/apiModels/TranscriptCoverageFilterResult.java
@@ -1,0 +1,27 @@
+package org.mskcc.cbio.oncokb.apiModels;
+
+public class TranscriptCoverageFilterResult {
+    String variant;
+    Boolean isCovered;
+
+    public TranscriptCoverageFilterResult(String variant, Boolean isCovered) {
+        this.variant = variant;
+        this.isCovered = isCovered;
+    }
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/apiModels/TranscriptCoverageFilterResult.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public String getVariant() {
+        return this.variant;
+    }
+
+    public void setVariant(String variant) {
+        this.variant = variant;
+    }
+
+    public Boolean isIsCovered() {
```
zhx828, line 20:
```suggestion
    public Boolean isCovered() {
```

zhx828 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/apiModels/TranscriptCoverageFilterResult.java
calvinlu3: Thanks for the catch. I actually used VSCode to generate getters/setters and it generated an extra getter. There's already the getter `getIsCovered`.

zhx828 (approved):

Issue #3417, opened 2023-03-16
Integrate clinical trials into IndicatorQueryResp


scarrero4660:
- [x] Add clinical trials as field in IndicatorQueryTreatment
- [x] Transfer logic to fetch trials to ClinicalTrialsUtils
- [x] Update IndicatorUtils to set trials in IndicatorTreatment, filtering by drugs and oncotree codes
- [x] Update clinical trial models
- [x] Add logic to fetch trials by main cancer type as well as by oncotree code

PR #3422, opened 2023-03-16
Integrate clinical trial data into IndicatorQueryResp for access in oncokbcard

scarrero4660:
This fixes https://github.com/oncokb/oncokb/issues/3417


scarrero4660:
After updating the models and IndicatorRespTreatment, running `yarn updateOncoKbAPI` did not update the corresponding types and fields in cbioportal-frontend's `OncoKbAPI.d.ts` file. I ensured that updateOncoKbAPI was using my local oncokbcore instance

core/src/main/java/org/mskcc/cbio/oncokb/model/clinicalTrialsMathcing/Trial.java
```diff
@@ -10,10 +10,15 @@
 public class Trial {
     private String briefTitle;
     private String currentTrialStatus;
+    private String currentTrialStatusDate;
```
zhx828, line 13: Do you think it makes more sense to convert it to a Date or Instant? Will we use it as a Date or simply just return it?

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -749,6 +750,23 @@ private static List<IndicatorQueryTreatment> getIndicatorQueryTreatments(Set<Evi
...
+                                    try {
+                                        List<Trial> trialsByTreatmentAndCode = clinicalTrialUtils.trialsMatchingGet(oncotreeCode,drugName);
+                                        trials.addAll(trialsByTreatmentAndCode);
+                                    } catch (Exception e) {
+                                        System.out.println(e.getMessage());
+                                        System.out.println("Trials could not be added successfully to IndicatorQueryTreatment.");
+                                    }
+                                }
+                                Set<Trial> trialsSet = new HashSet<>(trials);
+                                indicatorQueryTreatment.setTrials(trialsSet);
```
zhx828, line 768: We need to intersection of trials based on individual drug name. Trials that include all drugs should be returned.

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+import org.json.simple.JSONObject;
+import org.json.simple.parser.ParseException;
+import org.mskcc.cbio.oncokb.model.SpecialTumorType;
+import org.mskcc.cbio.oncokb.model.clinicalTrialsMathcing.*;
+import org.mskcc.cbio.oncokb.model.TumorType;
+
+public class ClinicalTrialsUtils {
+
+
+    public List<Trial> trialsMatchingGet(String oncoTreeCode, String treatment)
```
zhx828, line 20: We typically use format `getTrials`

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+                return new ArrayList<Trial>(tumor.getTrials());
+            }
+
+            List<Trial> trial = getTrialByTreatment(tumor.getTrials(), treatment);
+            return new ArrayList<Trial>(trial);
+        }
+        return new ArrayList<Trial>(new ArrayList<>());
+    }
+
+    public Map<String, List<Trial>> trialsGetByCancerTypes(CancerTypesQuery body)
```
zhx828, line 40: We typically use format `getTrialsByCancerTypes`

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+
+        JSONObject trialsJSON = CacheUtils.getTrialsJSON();
+        JSONObject oncotreeMappingJSON = CacheUtils.getOncoTreeMappingTrials();
+
+        Tumor tumor = new Tumor();
+        if (oncotreeMappingJSON.containsKey(oncoTreeCode)) {
+            tumor = getTumor(oncotreeMappingJSON,trialsJSON, oncoTreeCode);
+
+            if (treatment == null) {
+                return new ArrayList<Trial>(tumor.getTrials());
```
zhx828, line 31: You could do `new ArrayList<>(tumor.getTrials())`. No need to specify the type.

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private List<Trial> getTrialByTreatment(List<Trial> trials, String treatment) {
+        List<Trial> res = new ArrayList<>();
+        Set<String> drugsNames = Arrays.stream(treatment.split(",|\\+")).map(item -> item.trim()).collect(Collectors.toSet());
+
+        res = getTrialByDrugName(trials, drugsNames);
+        return res;
+    }
+
+    private List<Trial> getTrialByDrugName(List<Trial> trials, Set<String> drugsNames) {
```
zhx828, line 124:
```suggestion
    private List<Trial> getTrialsByDrugName(List<Trial> trials, Set<String> drugsNames) {
```

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+            for (Object t: trials) {
+                JSONObject trial = (JSONObject) t;
+                String nctID = (String) trial.get("nctId");
+                nctIDList.add(nctID);
+            }
+    
+            List<Trial> trialsInfo = new ArrayList<>();
+            for (String nctID: nctIDList) {
+                if (trialsJSON.containsKey(nctID)) {
+                    JSONObject trialObj = (JSONObject) trialsJSON.get(nctID);
```
zhx828, line 105: Can we cache the converted JsonObject instead? We need to convert to Trial anyway here.

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+                }
+                result.put(cancerType, addTrials);
+        }
+        return result;
+    }
+
+    private Tumor getTumor(JSONObject oncotreeMappingJSON, JSONObject trialsJSON, String oncoTreeCode) {
+        Tumor tumor = new Tumor();
+        if (oncotreeMappingJSON.containsKey(oncoTreeCode)) {
+            JSONObject tumorObj = (JSONObject) oncotreeMappingJSON.get(oncoTreeCode);
```
zhx828, line 90: Same goes here, we could just convert oncotreeMappingJSON to a Map<String, Tumor>

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,199 @@
...
+        TumorType matchedSpecialTumorType = ApplicationContextSingleton.getTumorTypeBo().getBySpecialTumor(specialTumorType);
+        if (matchedSpecialTumorType == null) return trials;
+
+        switch (specialTumorType) {
+            case ALL_TUMORS:
+                return new ArrayList<>(getAllTrials(tumors, trialData));
+            case ALL_SOLID_TUMORS:
+            case ALL_LIQUID_TUMORS:
+                return ApplicationContextSingleton.getTumorTypeBo().getAllTumorTypes().stream()
+                    .filter(tumorType -> tumorType.getTumorForm() != null && tumorType.getTumorForm().equals(matchedSpecialTumorType.getTumorForm()))
```
zhx828, line 156: if the tumorType has `MIXED` tumor form, it should be matched too.

zhx828 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/model/clinicalTrialsMathcing/Arm.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Created by Yifu Yao on 2020-09-08
  */
-public class Arms {
+public class Arm {
     private String armDescription;
```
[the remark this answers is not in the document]
scarrero4660: I believe `armDescription` is required because it has to match the schema outputted by the trials endpoint for the data to be returned. Otherwise null values will be returned

core/src/main/java/org/mskcc/cbio/oncokb/model/clinicalTrialsMathcing/Trial.java
scarrero4660: I don't believe this value is currently used or displayed anywhere but I had set it as a String because the format of the date in AACT was not consistent

core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
scarrero4660: @zhx828 can you clarify this? Here I am adding all of the trials that contain a drug name from the treatment list. Is that not what you are asking for?

zhx828 (approved):
LGTM, thanks!

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,201 @@
...
+                    }
+                }
+            }
+            tumor.setTrials(trialsInfo);
+        }
+
+        return tumor;
+    }
+
+    public static List<Trial> getTrialByTreatment(List<Trial> trials, String treatment) {
```
zhx828, line 126:
```suggestion
    public static List<Trial> getTrialsByTreatment(List<Trial> trials, String treatment) {
```

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
```diff
@@ -0,0 +1,201 @@
...
+        return new ArrayList<>(getAllTrialsBySpecialTumorType(specialTumorType));
+    }
+
+    public static List<Trial> getTrialsByCancerType(Map<String, Tumor>  oncotreeMapping, Map<String, Trial>  trialsMapping, String cancerType) {
+        List<Trial> trials = new ArrayList<>();
+
+        Set<String> tumorCodesByMainType = new HashSet<>();
+        List<TumorType> allOncoTreeSubtypes = ApplicationContextSingleton.getTumorTypeBo().getAllSubtypes();
+        for (TumorType oncoTreeType : allOncoTreeSubtypes) {
+            if (oncoTreeType.getMainType() != null && cancerType.equalsIgnoreCase(oncoTreeType.getMainType())) {
```
zhx828, line 168: cancerType could be null

zhx828 (changes requested):
Last few comments before merging

core/src/main/java/org/mskcc/cbio/oncokb/util/ClinicalTrialsUtils.java
scarrero4660: @zhx828 Just to clarify are you referring to the case when cancerType is null therefore we should return an empty list or when oncoTreeType.getMainType() is null?

zhx828 (approved):
LGTM

PR #3425, opened 2023-03-21
Replace putIfAbsent with put when caching pre-annotated gn variants

calvinlu3:
This is related to https://github.com/oncokb/oncokb-pipeline/issues/107

The pre-annotated GN variants can be cached at a rate of 100K/20s if we use the `cache.put()` method instead  `cache.putIfAbsent()` method.

With a total of ~2.2 million variants, we can cache all in about 8 mins (~1min for high priority and 7mins for low priority)


zhx828 (changes requested):
Could you do me a favor to update the signature of method `getAlterationFromGenomeNexus` in AlterationUtils from `(GNVariantAnnotationType type, String query, ReferenceGenome referenceGenome)` to `(GNVariantAnnotationType type, ReferenceGenome referenceGenome, String query)` so it aligns with the method in the CacheFetcher?

zhx828 (approved):

PR #3432, opened 2023-04-04
Improve performance of annotate by hgvsg and gc endpoints

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3431

## Changes
- The main bottleneck was calling `cacheFetcher.getAllTranscriptGenes()` for each query inside the post body. For post requests, this call is moved outside the for loop, so it is only called once per request.
- A smaller improvement was using java stream's `anyMatch()` instead of `filter().count() > 0`, to when determining if the gc/hgvsg should be annotated by GN. `anyMatch()` will return when the first element is True, whereas filter will continue.

## Performance comparison
- 1 oncokb-core pod
- 1 thread, iterating 5 times and taking average
- POST request with 500 HGVSg, [test-post-body.zip](https://github.com/oncokb/oncokb/files/11151835/test-post-body.zip)

Original implementation: 5.4s
Moving `cacheFetcher.getAllTranscriptGenes()` outside for loop:  1.7s
Using `anyMatch()`: 1.2s


zhx828 (approved):
LGTM, thanks!

PR #3433, opened 2023-04-04
Make abbreviation searchable & Support PTD in abbreviation

zhx828:


core/src/main/resources/data/abbreviation-ontology.tsv
```diff
@@ -1,9 +1,10 @@
-# Abbreviation	Full Name
+# Abbreviation(will be converted to all uppercase)	Full Name(This is preferred name that used in the summary.)
```
calvinlu3, line 1:
```suggestion
# Abbreviation(will be converted to all uppercase)	Full Name(This is preferred name that is used in the summary.)
```

calvinlu3 (approved):

PR #3445, opened 2023-04-18
Use original query for GN cache endpoint

calvinlu3:
Changes:
- If original query is HGVSg, then use the original query for `hgvsg` field. Same for genomic location.
- In `getConsequence` method, we should search for gene by entrezId, hugoSymbol, gene alias. In older Ensembl API, they may return a hugoSymbol that is an alias.


zhx828 (approved):
LGTM

Issue #3462, opened 2023-06-01
Mockup to include GN/Revue logo in the HGVSg page


zhx828:
Latest mockup: https://www.figma.com/file/PKwXQzD5e80UWMGFw94Hdw/hgvsg-page?type=design&node-id=0-1&t=ThIenn0BlDb2oAfc-0

PR #3482, opened 2023-06-23
Include VUE info from genome nexus

zhx828:
This is related to https://github.com/oncokb/oncokb/issues/3462
We want to show reVUE logo when searching HGVSg and it's annotated by reVUE.

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
@@ -2,6 +2,7 @@ package org.mskcc.cbio.oncokb.api.pvt;
 
 import com.mysql.jdbc.StringUtils;
 import io.swagger.annotations.ApiParam;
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
```diff
@@ -410,6 +411,14 @@ public ResponseEntity<VariantAnnotation> utilVariantAnnotationGet(
 
         VariantAnnotation annotation = new VariantAnnotation(indicatorQueryResp);
 
+        // for any hgvsg variant, we need to check whether it is VUE
+        if(!StringUtils.isNullOrEmpty(hgvsg)) {
+            TranscriptConsequenceSummary transcriptConsequenceSummary = GenomeNexusUtils.getTranscriptConsequence(GNVariantAnnotationType.HGVS_G, hgvsg, matchedRG);
```
calvinlu3, line 416: Also check if `transcriptConsequenceSummary` is null?

calvinlu3 (approved):

PR #3488, opened 2023-07-05
Add an option to send genomicChange through variantAnnotation endpoint

zhx828:
This is related to https://github.com/oncokb/oncokb/issues/3464

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
@@ -363,6 +363,7 @@ public class PrivateUtilsApiController implements PrivateUtilsApi {
         , @ApiParam(value = "Reference genome, either GRCh37 or GRCh38. The default is GRCh37", defaultValue = "GRCh37") @RequestParam(value = "referenceGenome", required = false, defaultValue = "GRCh37") String referenceGenome
         , @ApiParam(value = "Alteration") @RequestParam(value = "alteration", required = false) String alteration
         , @ApiParam(value = "HGVS genomic format. Example: 7:g.140453136A>T") @RequestParam(value = "hgvsg", required = false) String hgvsg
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
```diff
@@ -377,7 +378,18 @@ public ResponseEntity<VariantAnnotation> utilVariantAnnotationGet(
...
+        if (!StringUtils.isNullOrEmpty(hgvsg) || !StringUtils.isNullOrEmpty(genomicChange)) {
+            String genomicQuery = StringUtils.isNullOrEmpty(hgvsg) ? genomicChange : hgvsg;
+            GNVariantAnnotationType type = StringUtils.isNullOrEmpty(hgvsg) ? GNVariantAnnotationType.GENOMIC_LOCATION : GNVariantAnnotationType.HGVS_G;
+            Alteration alterationModel;
+            if (!this.cacheFetcher.genomicLocationShouldBeAnnotated(type, genomicQuery, matchedRG, this.cacheFetcher.getAllTranscriptGenes())) {
+                alterationModel = new Alteration();
+            } else {
+                alterationModel = this.cacheFetcher.getAlterationFromGenomeNexus(type, matchedRG, genomicQuery);
+            }
+            query = QueryUtils.getQueryFromAlteration(matchedRG, tumorType, alterationModel, hgvsg);
```
calvinlu3, line 390: Just a quick question, why do we save the HGVSg into the Query object and not the genomic change?

calvinlu3 (approved):
LGTM

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
zhx828: It's a historical issue that I didn't include when rolling out the byGenomicChange annotation endpoint. I don't intend to fix it in this PR as the changes are related to a private endpoint. I usually don't change models that used in the published endpoints.

PR #3489, opened 2023-07-10
Update tests for data 4.6

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3477


core/src/test/resources/test_tumor_type_summaries.tsv
```diff
@@ -158,19 +158,19 @@ ERBB2	fusion	Breast cancer	ERBB2, a receptor tyrosine kinase, is altered by muta
...
+FGFR2	KIF14-FGFR2 fusion	Cancer of Unknown Primary	FGFR2, a receptor tyrosine kinase, is altered by mutation, chromosomal rearrangement or amplification in various cancer types.	The KIF14-FGFR2 fusion is likely oncogenic.	While the pan-FGFR-targeted inhibitors pemigatinib and futibatinib are FDA-approved for the treatment of patients with FGFR2-fusion positive cholangiocarcinoma, and erdafitnib is FDA-approved for the treatment of patients with urothelial cancer with susceptible FGFR genetic alterations, their clinical utility in patients with KIF14-FGFR2 fusion positive cancer of unknown primary is unknown.
 FGFR3	K650E	Glioma	FGFR3, a receptor tyrosine kinase, is altered by mutation, chromosomal rearrangement or amplification in various cancers, most frequently in bladder cancer.	The FGFR3 K650E mutation is known to be oncogenic.	While there is promising clinical data in patients with urothelial cancer harboring functionally characterized hotspot FGFR3 mutations treated with pan-FGFR-targeted inhibitors such as AZD4547, BGJ398, Debio1347 and erdafitinib, their clinical utility in patients with FGFR3 K650E mutant glioma is unknown.
```
zhx828, line 174: Seems like the summery under FGFR3/S371C, G380R, K650/Other Tumor Types is not updated by remove Debio1347. Probably good to check all descriptions whether `Debio1347` is still mentioned anywhere. May as well check `Infigratinib`, it was removed this May.

zhx828 (changes requested):

core/src/test/resources/test_tumor_type_summaries.tsv
calvinlu3: What is the best way to remove `Debio1347` from description? Should I edit on the curation platform and ask for review, so it triggers a history save?

zhx828: @calvinlu3 could you look into the beta db and see which sections still have `Debio1347`? Then let the curation team know where they are, they should fix.

zhx828 (approved):
LGTM, thank you.

PR #3513, opened 2023-07-28
Sort gene/alteration bases on prevalence data for cancer type

zhx828:
The gene with highest occurrence within the prevalence data will be shown first in the cancer type search.
This is related to https://github.com/oncokb/oncokb/issues/3508

Screenshot 2023-07-28 at 6 52 35 PM


web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateSearchApiController.java
```diff
@@ -757,3 +808,44 @@ public int compare(LevelOfEvidence o1, LevelOfEvidence o2) {
     }
 }
 
+class GeneCount {
```
Eric-Jiang1, line 811: Do we want this in a separate file?

Eric-Jiang1 (approved):

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateSearchApiController.java
zhx828: Not necessarily if they are not used anywhere else.

Issue #3490, opened 2023-07-11
Test whether any evidence contains a drug not in db


calvinlu3:
We should add a validation test to see if any of the evidence descriptions contain a drug that is no longer in our system.

PR #3523, opened 2023-08-04
Add a test to check whether evidence includes outdated content

zhx828:
The first check is to find removed drug from the description.

This fixes https://github.com/oncokb/oncokb/issues/3490

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/ValidationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/ValidationUtils.java
@@ -8,6 +8,7 @@ import com.mysql.jdbc.StringUtils;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.mskcc.cbio.oncokb.bo.ArticleBo;
+import org.mskcc.cbio.oncokb.bo.DrugBo;
 import org.mskcc.cbio.oncokb.model.*;
 
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/ValidationUtils.java
```diff
@@ -134,6 +134,42 @@ public static JSONArray checkEvidenceDescriptionReferenceFormat() {
...
+        List<Drug> allDrugs = drugBo.findAll();
+        Set<Drug> drugsWithoutEviAssoc = allDrugs.stream().filter(drug -> !drugsWithEviAssoc.contains(drug)).collect(Collectors.toSet());
+
+        Set<String> drugNameWithSynonyms = new HashSet<>();
+        for (Drug drug : drugsWithoutEviAssoc) {
+            drugNameWithSynonyms.addAll(drug.getSynonyms());
+            drugNameWithSynonyms.add(drug.getDrugName());
+        }
+        Set<EvidenceType> evidenceTypesToCheck = new HashSet<>();
+        evidenceTypesToCheck.add(EvidenceType.MUTATION_EFFECT);
```
calvinlu3, line 156: If I remember correctly, it is fine to mention a deleted drug in the mutation effect summary. 
Do you think we should still include this or its valuable to let the curators know about these summaries?

zhx828: @calvinlu3 Good point, ME is tricky. Let's set it up for now. We will do one data release, then decide whether to exclude.

calvinlu3 (approved):

Issue #3554, opened 2023-09-05
Inframe-insertion is interpreted as splice variant.


zhx828:
https://www.genomenexus.org/variant/7:g.116411902_116411903insAAT
https://www.oncokb.org/genomic-change/7,116411902,116411902,G,GAAT
Three codons were inserted right after the splice site, GN gives splice region variant as consequence which leads to annotation using the splice mutation information. However the variant class from GN seems correctly annotate it as inframe-insertion.

<img width="1622" alt="Screenshot 2023-09-05 at 5 23 04 PM" src="https://github.com/oncokb/oncokb/assets/5400599/4dd21551-53b1-4666-83a8-eabcf91a0668">

PR #3558, opened 2023-09-08
Add variant consequence resolver

zhx828:
When consequence is specified by user or by GenomeNexus, there are situations that
- there are multiple consequences associated with the variant
- conflict with variant classification

This resolver fixes https://github.com/oncokb/oncokb/issues/3554

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/VariantConsequenceUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/VariantConsequenceUtils.java
@@ -107,12 +107,9 @@ public class VariantConsequenceUtils {
         if (consequenceTerms == null) {
             consequenceTerms = "";
         }
-        if (variantClass == null) {
-            variantClass = "";
-        }
 
         VariantConsequence consequence = findVariantConsequenceByTerm(consequenceTerms.contains(",") ? pickConsequenceTerm(consequenceTerms) : consequenceTerms);
-        VariantConsequence variantClassConsequence = variantClass == null ? null : findVariantConsequenceByTerm(variantClass);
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/VariantConsequenceUtils.java
```diff
@@ -94,6 +95,46 @@ private static void cacheVariantConsequencesMap() {
...
+        }
+        if (consequenceTerms == null) {
+            consequenceTerms = "";
+        }
+        if (variantClass == null) {
+            variantClass = "";
+        }
+
+        VariantConsequence consequence = findVariantConsequenceByTerm(consequenceTerms.contains(",") ? pickConsequenceTerm(consequenceTerms) : consequenceTerms);
+        VariantConsequence variantClassConsequence = variantClass == null ? null : findVariantConsequenceByTerm(variantClass);
```
calvinlu3, line 115: `variantClass` will never be null since on lines 110-112, it is assigned to empty string.

core/src/main/java/org/mskcc/cbio/oncokb/util/VariantConsequenceUtils.java
```diff
@@ -94,6 +95,46 @@ private static void cacheVariantConsequencesMap() {
...
+
+        if (consequence == null) {
+            if (variantClassConsequence != null) {
+                return variantClassConsequence;
+            } else {
+                return null;
+            }
+        } else {
+            if (variantClassConsequence != null && variantClassConsequence != consequence) {
```
calvinlu3, line 124:
```suggestion
            if (variantClassConsequence != null && variantClassConsequence.equals(consequence)) {
```
Should we use the `equals` method that is defined in `VariantConsequence` class? `!=` will just be comparing the references to the two object.

calvinlu3 (approved):

PR #3618, opened 2023-12-01
Update tests for data v4.12

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3615


core/src/test/resources/test_mutation_effect_summaries.tsv
```diff
@@ -28,7 +28,7 @@ BRAF	T100*	Unknown
...
 ### Truncating mutations in TSG that not curated
-PPP2R2A	T100*	Unknown
+PPP2R2A	T100*	Likely Loss-of-function	The mutation effect description for truncating mutations in PPP2R2A is: PPP2R2A truncating mutations typically result in the loss of protein function. The loss of PPP2R2A activity results in reduced PP2A phosphatase activity and altered activity of PP2A target proteins (PMID: 27531894). In addition, reduced PPP2R2A activity has been associated with altered DNA damage responses and increased sensitivity to DNA damaging agents (PMID: 32522823, 23087057).
```
zhx828, line 31: We need to find a gene that is TSG but Truncating mutations is not curated.

zhx828 (approved):
The main concern is related to the TSG Truncating Mutations. If we can not find a gene that without Truncating Mutations, we need to redesign.

Issue #3612, opened 2023-11-22
Throw exception when genome-nexus timeout


zhx828:
To be validated:

User mentioned during a period of time, for the same coordinates, OncoKB returns Unknown. One potential possibility is GN timedout. In this case, we should throw exception. For both POST and GET

zhx828:
Tested the timeout exception which oncokb-core will throw when it happens. Also checked the redis which will not save any data when there is an exception.

PR #3635, opened 2023-12-18
Do not capture GN ApiException in POST method

zhx828:
This is related to https://github.com/oncokb/oncokb/issues/3612


calvinlu3 (approved):
Looks good

PR #3642, opened 2023-12-21
Add annotation search endpoint

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/209

### Feature
Create a search endpoint that utilizes our IndicatorQuery.processQuery() logic to get annotations for search query.
For single keyword queries, we will return matches for Gene, Alteration and Cancer Type
For multi-keyword queries, we are only able to support Gene + Alteration, Alteration, and Cancer Type.

### Limitations
1. For multiple keyword queries, we should be able to not assume that the first keyword is a Gene.
2. We are not able to support searches that have gene, alteration and cancer types.
3. We don't support drug searches, but should be a simple addition


calvinlu3:
@zhx828 
For `Pineoblastoma`, the search endpoint returns no results because all the therapeutics are found under `All Solid Tumors`. Based on our discussion last time, we didn't want to include parent cancer types.

zhx828:
Few issues while testing. Let me know if you want me to fix the following.

- [ ] when alteration is not available, the variant summary should not be returned
- [ ] when search BRAF, only three results are returned(one gene, two variants). But the current search bar shows more.
- [ ] when search BRAF V600E, the second records seems related to `p61braf-V600E`, but the query shows `V169_D380del`
- [ ] when search `BCL6 Fusions`, it does not return any result. But it should.
- [ ] The cancer type search result should be sorted by cancer type, then level
- [ ] The cancer type search(melanoma) only returns 31 records, but actionable genes page shows 60
- [ ] Drug search should be supported

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateAnnotationController.java
```diff
@@ -0,0 +1,353 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+import java.util.Objects;
+
+@RestController
+@Api(tags = "Annotation")
```
zhx828, line 39: This should be Annotations so it can be included with other `annotate` endpoints.

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateAnnotationController.java
```diff
@@ -0,0 +1,353 @@
...
+public class PrivateAnnotationController {
+
+    @PremiumPublicApi
+    @ApiOperation(value = "", notes = "Get annotations based on search", response = AnnotationSearchResult.class, responseContainer = "List")
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "OK")})
+    @RequestMapping(value = "/annotation/search",
```
zhx828, line 46: This endpoint is currently under `api/v1` I assume this is because of the @ PremiumPublicApi. Let's move this to pub/v1

zhx828 (commented):
I didn't review the code in details. Will do it after some of the comments are addressed.

calvinlu3:
`The cancer type search(melanoma) only returns 31 records, but actionable genes page shows 60`
- I tested using Ocular Melanoma, which has 43 records on actionable genes page. Since search results are grouped by gene, alteration and cancer type, we should expect 39 records.

@zhx828 I did not implement drug search because I would probably implement it in a similar way as the cancer type search (which needs to be reviewed). I'm thinking that I need to get all evidences with that drug and then construct `SearchObjects` to pass to `processQuery()`

zhx828:
@calvinlu3 `when search BRAF V600E, the second records seems related to p61braf-V600E, but the query shows V169_D380del` this seems to be still the case?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -517,7 +519,186 @@ public ResponseEntity<List<IndicatorQueryResp>> annotateStructuralVariantsPost(
...
+            return new LinkedHashSet<>();
+        }
+
+        LinkedHashSet<AnnotationSearchResult> result = new LinkedHashSet<>();
+        Set<SearchObject> searchObjects = new HashSet<>();
+        for (TumorType tumorType : tumorTypeMatches) {
+            for (Evidence evidence : allImplicationEvidences) {
+                if (evidence.getGene().getHugoSymbol().equals("NTRK3")) {
+                    Boolean t = true;
+                }
```
zhx828, line 671: What's this?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -600,3 +781,152 @@ private static Gene findGeneFromPool(Map<String, Gene> genePool, QueryGene query
...
+            if (a1.getQueryType().equals(AnnotationSearchQueryType.VARIANT)) {
+                name1 = i1.getQuery().getAlteration().toLowerCase();
+                name2 = i2.getQuery().getAlteration().toLowerCase();
+            }
+            if (a1.getQueryType().equals(AnnotationSearchQueryType.CANCER_TYPE)) {
+                name1 = i1.getQuery().getTumorType().toLowerCase();
+                name2 = i2.getQuery().getTumorType().toLowerCase();
+                if (name1.contains("acral") || name2.contains("acral")) {
+                    Boolean t = true;
+                }
```
zhx828, line 816: What's this?

calvinlu3:
@zhx828 Sorry about that, I had it fixed for alteration search, but the change didn't reflect for gene+alteration. Fixed now.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
calvinlu3: Thanks for the catch. I was using it to debug some scenarios. Removed.

PR #3662, opened 2024-02-02
Update test for v4.14

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3661


core/src/test/java/org/mskcc/cbio/oncokb/util/FindRelevantAlterationsTest.java
```diff
@@ -105,11 +105,11 @@ public static Collection<String[]> getParameters() {
 //                {"MAP2K1", "N109_R113del", null, "N109_R113del, I99_R113del"},
 
                 // Range missense variant
-                {"PDGFRA", "D842I", null, "D842I, D842H, D842Y, D842_I843delinsIM, 814_852mis, Oncogenic Mutations"},
```
zhx828, line 108: The D842H is still not included?

PR #3664, opened 2024-02-05
Fix excluding issue for mutation and cancer type

zhx828:


calvinlu3 (approved):
This makes sense, LGTM

PR #3673, opened 2024-02-14
Add annotation search endpoint to pvt api

calvinlu3:
Duplicating the annotation search endpoint to allow time for migration


calvinlu3:
Not needed

Issue #3684, opened 2024-02-26
Update the error message to include more details


zhx828:
When oncokb-core API fail, the error message only showing below which is not informative at all.

{
  "title": "400 : [no body]",
  "status": 400,
  "message": "error.",
  "params": ""
}

Example:
Making API call to `/annotate/mutations/byHGVSg` with `hgvsg=8g.67488496C%3EA`.

jfkonecn:
@zhx828 

On a bad request the byHGVSg endpoint isn't returning a body at all.

Looks like all the endpoints in the [AnnotationsApiController](https://github.com/oncokb/oncokb/blob/master/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java) do the same thing so I think we have two options.

1. Fix all the endpoints in the controller so they have a better error messages.
2. Create a generic error message in OncoKB Public

What do you want me to do?

```java
    @PublicApi
    @PremiumPublicApi
    @ApiOperation(value = "", notes = "Annotate mutation by HGVSg.", response = IndicatorQueryResp.class)
    @ApiResponses(value = {
        @ApiResponse(code = 200, message = "OK", response = IndicatorQueryResp.class),
        @ApiResponse(code = 400, message = "Error, error message will be given.", response = String.class)})
    @RequestMapping(value = "/annotate/mutations/byHGVSg",
        produces = {"application/json"},
        method = RequestMethod.GET)
    public ResponseEntity<IndicatorQueryResp> annotateMutationsByHGVSgGet(
        @ApiParam(value = "HGVS genomic format. Example: 7:g.140453136A>T", required = true) @RequestParam(value = "hgvsg", required = true) String hgvsg
        , @ApiParam(value = "Reference genome, either GRCh37 or GRCh38. The default is GRCh37", required = false, defaultValue = "GRCh37") @RequestParam(value = "referenceGenome", required = false, defaultValue = "GRCh37") String referenceGenome
        , @ApiParam(value = "OncoTree(http://oncotree.info) tumor type name. The field supports OncoTree Code, OncoTree Name and OncoTree Main type. Example: Melanoma") @RequestParam(value = "tumorType", required = false) String tumorType
        , @ApiParam(value = EVIDENCE_TYPES_DESCRIPTION) @RequestParam(value = "evidenceType", required = false) String evidenceTypes
    ) throws ApiException, org.genome_nexus.ApiException {
        HttpStatus status = HttpStatus.OK;
[...]

PR #3707, opened 2024-03-07
Add error messages for annotations api controller

jfkonecn:
#3684

This is how the error looks coming out of public

```json
{
    "title": "400 : [Please provide a valid HGVS genomic format.]",
    "status": 400,
    "message": "error.",
    "params": ""
}
```

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
@@ -44,7 +44,7 @@ public class AnnotationsApiController {
     @RequestMapping(value = "/annotate/mutations/byProteinChange",
         produces = {"application/json"},
         method = RequestMethod.GET)
-    public ResponseEntity<IndicatorQueryResp> annotateMutationsByProteinChangeGet(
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -44,7 +44,7 @@ public class AnnotationsApiController {
     @RequestMapping(value = "/annotate/mutations/byProteinChange",
         produces = {"application/json"},
         method = RequestMethod.GET)
-    public ResponseEntity<IndicatorQueryResp> annotateMutationsByProteinChangeGet(
+    public ResponseEntity<Object> annotateMutationsByProteinChangeGet(
```
zhx828, line 47: We can't change this. The API client will not be able to generate expected response body.

jfkonecn: Doesn't this existing attribute take care of that problem?

```java
@ApiResponse(code = 400, message = "Error, error message will be given.", response = String.class)})
```

jfkonecn:
Will use ControllerAdvice instead

PR #3714, opened 2024-03-20
Expose endpoint to return all variants of unknown significance

zhx828:
These two endpoints do not have version as parameter. We don't have vus files listed in previous versions, and we don't have intention to maintain that.

This fixes https://github.com/oncokb/oncokb-pipeline/issues/291


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -93,6 +94,35 @@ public ResponseEntity<String> utilsAllAnnotatedVariantsTxtGet(
         return new ResponseEntity<>(sb.toString(), HttpStatus.OK);
     }
 
+    @Override
+    public ResponseEntity<List<VariantOfUnknownSignificance>> utilsAllVariantsOfUnknownSignificanceGet() {
+        return new ResponseEntity<>(getAllVus(), HttpStatus.OK);
+    }
+
+    @Override
+    public ResponseEntity<String> utilsAllVariantsOfUnknownSignificanceTxGet() {
```
calvinlu3, line 103:
```suggestion
    public ResponseEntity<String> utilsAllVariantsOfUnknownSignificanceTxtGet() {
```

calvinlu3 (approved):
LGTM

PR #3715, opened 2024-03-21
Standardize the text files headers

zhx828:
- The PMIDs/Abstract/Description are for both mutation effect and oncogenicity;
- The PMIDs/Abstract/Description are for Tx with drug and also for Dx/Px without drug;


calvinlu3 (approved):
LGTM

PR #3726, opened 2024-04-10
Added error message to 404 error for utils controller

jfkonecn:
The byte[] was giving me issues when I tried to change the type to ResponseEntity<byte[]> so I instead switched it to ResponseEntity<?>. 

I did some further reading an it seems like the proper way to do this is by using a "@ControllerAdvice"
[stack overflow question about returning success or error](https://stackoverflow.com/questions/38117717/what-is-the-best-way-to-return-different-types-of-responseentity-in-spring-boot)
[Article about @ControllerAdvice](https://www.toptal.com/java/spring-boot-rest-api-error-handling)
[Spring blog post about @ControllerAdvice](https://spring.io/blog/2013/11/01/exception-handling-in-spring-mvc)

I feel like using a @ControllerAdvice is a bigger change than what you were looking for, but let me know what you want me to do.


jfkonecn:
Will use ControllerAdvice instead

Issue #3724, opened 2024-04-04
Stop loss should not be considered as truncating mutation


zhx828:
User reported with `JAK1 *1155Qext*3	a stoploss mutation should not be annotated as truncating`

zhx828:
From Sarah `I have not really come across this before, but basically the stop codon is lost and the protein is extended beyond the stop codon. So the user is right, it is not a truncating event (edited) `

zhx828:
We have been annotating stop_lost as truncating mutation since our inception. Will update after team discussion

zhx828:
Update the consequence to false on is_generally_truncating

PR #3740, opened 2024-05-29
Add a test on stop_lost relevant alts

zhx828:
The stop_lost should not be considered as truncating mutation therefore the test should not have `Truncating Mutations, Oncogenic Mutations` returned.

This is related to https://github.com/oncokb/oncokb/issues/3724


calvinlu3 (approved):
LGTM.

Just a quick question: How is `is_generally_truncating` determined? I remember going to the ensembl website that is linked at the top of the txt file, but was not sure how we populated that field.

PR #3751, opened 2024-07-02
Add validation test for actionable variants and oncogenicity

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/462

image

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
@@ -15,6 +15,7 @@ import java.io.IOException;
 import java.util.*;
 import java.util.stream.Collectors;
 
+import static org.mskcc.cbio.oncokb.Constants.DEFAULT_REFERENCE_GENOME;
 import static org.mskcc.cbio.oncokb.api.websocket.ValidationCategory.*;
 import static org.mskcc.cbio.oncokb.model.StructuralAlteration.TRUNCATING_MUTATIONS;
 
@@ -24,6 +25,7 @@ import static org.mskcc.cbio.oncokb.model.StructuralAlteration.TRUNCATING_MUTATI
 
 @ServerEndpoint(value = "/api/websocket/curation/validation")
 public class CurationValidationApiController {
+
     private Session session;
 
     @OnOpen
@@ -50,6 +52,8 @@ public class CurationValidationApiController {
         validateTruncatingMutationsUnderTSG();
 
         validateMismatchedRefAA();
+
+        validateVariantActionabilityAndOncogenicity();
 
         compareActionableGenes();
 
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
```diff
@@ -341,6 +363,42 @@ public JSONArray getTruncatingMutationsNotUnderTSG() throws ApiException {
...
+            if (!allowedOncogenicities.contains(response.getOncogenic())) {
+                if (response.getHighestSensitiveLevel() != null || response.getHighestResistanceLevel() != null) {
+                    String hugoSymbol = alteration.getGene().getHugoSymbol();
+                    StringBuilder messageBuilder = new StringBuilder();
+                    messageBuilder.append(hugoSymbol);
+                    messageBuilder.append(" ");
+                    messageBuilder.append(alteration.getAlteration());
+                    messageBuilder.append(" is ");
+                    messageBuilder.append(response.getOncogenic());
+                    messageBuilder.append(", but is actionable.");
```
zhx828, line 394: Can we provide some additional info in terms of what is the actionable here? Maybe include the drug and level?

calvinlu3: My thought was that there could be multiple CTs and Treatments under a mutation, which could result in a long text. I imagine the mutation + alteration would be enough to point the curators to the right place, so they can delete all treatments.

calvinlu3:
image

web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
```diff
@@ -1,6 +1,6 @@
 package org.mskcc.cbio.oncokb.api.websocket;
 
-import com.mysql.jdbc.StringUtils;
```
calvinlu3, line 3: Noticed that we use StringUtils from mysql instead of apache in a couple of files.

web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
```diff
@@ -341,8 +363,59 @@ public JSONArray getTruncatingMutationsNotUnderTSG() throws ApiException {
...
+                            StringBuilder errorMessage = new StringBuilder();
+                            errorMessage.append("Is ");
+                            errorMessage.append(response.getOncogenic());
+                            errorMessage.append(", but has ");
+                            errorMessage.append(treatment.getLevel().toString());
+                            errorMessage.append(" treatment: ");
+                            List<String> drugList = treatment.getDrugs().stream().map(drug -> {
+                                return drug.getDrugName();
+                            }).collect(Collectors.toList());
+                            errorMessage.append(StringUtils.join(drugList, " + "));
```
zhx828, line 406: Essentially when the variant has multiple treatments, the `is inconclusive` message will be repeated?

web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
zhx828: Thanks, we can migrate when possible.

web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
calvinlu3: Yeah it could be possible `Gene / Alteration / CT` is repeated if there is are multiple treatments.

zhx828 (approved):
Thanks

PR #3759, opened 2024-07-17
Improve frameshift variant parse

zhx828:
Recognize frameshift and compare when ref/var AA are not available and when extension is different.
This fixes https://github.com/oncokb/oncokb-pipeline/issues/472


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/AlterationBoImpl.java
```diff
@@ -30,99 +30,6 @@ public List<Alteration> findAlterationsByGene(Collection<Gene> genes) {
         return alterations;
     }
 
-    public Alteration findExactlyMatchedAlteration(ReferenceGenome referenceGenome, Alteration alteration, List<Alteration> fullAlterations) {
```
zhx828, line 33: Moved the method to AlterationUtils.

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -1327,6 +1370,141 @@ private static List<Alteration> findAlterationsByStartWith(String startWith, Lis
         return matchedAlterations.stream().collect(Collectors.toList());
     }
 
+    public static Alteration findExactlyMatchedAlteration(ReferenceGenome referenceGenome, Alteration alteration, List<Alteration> fullAlterations) {
```
zhx828, line 1373: The following methods are from AlterationBoImpl. But please review lines 1416-1426.

calvinlu3 (approved):
LGTM

PR #3762, opened 2024-07-19
Fix tumor type summary mapping

zhx828:
- fix a bug that evidence been deleted from matching
- sort evidence based on number of alterations associated. Evidence with fewer alterations associated is put to the front

This fixes https://github.com/oncokb/oncokb-pipeline/issues/480


core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/EvidenceBoImpl.java
```diff
@@ -95,18 +99,17 @@ public List<Evidence> findEvidencesByAlteration(Collection<Alteration> alteratio
...
         }
 
-        List<Evidence> sortedEvidences = new ArrayList<>();
+        Set<Evidence> sortedEvidences = new LinkedHashSet<>();
         for (TumorType tumorType : relevantMatchedTumorTypes) {
             for (int i = 0; i < evidences.size(); i++) {
                 Evidence evidence = evidences.get(i);
                 if (evidence.getCancerTypes().contains(tumorType)) {
                     sortedEvidences.add(evidence);
-                    evidences.remove(evidence);
```
calvinlu3, line 104: So basically, this line is messing the index in the loop

calvinlu3 (approved):
LGTM

core/src/main/java/org/mskcc/cbio/oncokb/bo/impl/EvidenceBoImpl.java
zhx828: yup

Issue #3758, opened 2024-07-17
*1069Ffs*5 should not be considered as truncating mutation


zhx828:
When a frameshift happens at the stop codon, we should not consider it as truncating mutation. The original protein chain is extended rather than truncated.
However if the format is `*1069fs*?`, that means the new frame does not have a stop codon formed, it's unknown how many AA are extended. We should not consider `*1069fs*?` to be the same as `*1069fs*5`

PR #3764, opened 2024-07-22
Stop codon frameshift alt should not be considered as truncating

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3758


calvinlu3 (approved):
LGTM

PR #3768, opened 2024-08-01
Improve abstract regex

zhx828:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/468

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/ArticleUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/ArticleUtils.java
@@ -1,0 +1,54 @@
+package org.mskcc.cbio.oncokb.util;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/ArticleUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.LinkedHashSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+import org.mskcc.cbio.oncokb.model.Article;
+
+public class ArticleUtils {
+    public static LinkedHashSet<Article> getAbstractFromText(String text) {
+        LinkedHashSet<Article> articles = new LinkedHashSet<>();
+        if (text == null) return articles;
```
calvinlu3, line 13: Minor, but we can return early if empty string too.
```suggestion
        if (StringUtils.isEmpty(text)) return articles;
```

PR #3771, opened 2024-08-06
Added controller advice for Annotations API Controller

jfkonecn:
Closes oncokb#3684

I focused on just `AnnotationsApiController.java` to see if you think I'm heading in the right direction. I can expand this to other controllers if you want me to.

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ApiHttpError.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ApiHttpError.java
@@ -2,17 +2,47 @@ package org.mskcc.cbio.oncokb.controller.advice;
 
 import org.springframework.http.HttpStatus;
 
+/**
+* Represents an HTTP error response with detailed information about the error.
+* This class captures the title, status, detail, path, and message associated with the HTTP error.
+*
+* Example JSON representation:
+* <pre>
+* {
+*   "type": "https://www.oncokb.org/problem/problem-with-message",
+*   "title": "Method Not Allowed",
+*   "status": 405,
+*   "detail": "Request method 'POST' is not supported",
+*   "instance": "/index.html",
+*   "message": "error.http.405",
+*   "path": "/index.html"
+* }
+* </pre>
+*/
 public class ApiHttpError {
-    private String type;
+    /** The title of the error, typically representing the HTTP status text. */
     private String title;
+
+    /** The HTTP status code of the error (e.g., 404, 500). */
     private int status;
+
+    /** A detailed message explaining the error. */
     private String detail;
+
+    /** The request path that caused the error. */
     private String path;
+
+    /** A message key that is standardize to help with client side error handling, such as "error.http.404". */
     private String message;
 
+    /**
+    * Constructs an ApiHttpError object with the specified HTTP path, status, and message.
+    *
+    * @param path    The request path that caused the error.
+    * @param status  The HTTP status code of the error.
+    * @param message A detailed message explaining the error.
+    */
     public ApiHttpError(String path, HttpStatus status, String message) {
-        // TODO: add problem endpoint
-        // this.setType("https://www.oncokb.org/problem/problem-with-message");
         this.setPath(path);
         this.setTitle(status.toString());
         this.setStatus(status.value());
@@ -20,50 +50,92 @@ public class ApiHttpError {
         this.setMessage("error.http." + status.value());
     }
 
-    public String getType() {
-        return type;
-    }
-
-    public void setType(String type) {
-        this.type = type;
-    }
-
+    /**
+    * Gets the title of the error, typically representing the status text.
+    *
+    * @return The title of the error.
+     */
     public String getTitle() {
         return title;
     }
 
+    /**
+    * Sets the title of the error.
+    *
+    * @param title The title of the error.
+     */
     public void setTitle(String title) {
         this.title = title;
     }
 
+    /**
+    * Gets the HTTP status code of the error.
+    *
+    * @return The HTTP status code.
+     */
     public int getStatus() {
         return status;
     }
 
+    /**
+    * Sets the HTTP status code of the error.
+    *
+    * @param status The HTTP status code.
+     */
     public void setStatus(int status) {
         this.status = status;
     }
 
+    /**
+    * Gets the detailed description of the error.
+    *
+    * @return The detail of the error.
+     */
     public String getDetail() {
         return detail;
     }
 
+    /**
+    * Sets the detailed description of the error.
+    *
+    * @param detail The detail of the error.
+     */
     public void setDetail(String detail) {
         this.detail = detail;
     }
 
+    /**
+    * Gets the request path that caused the error.
+    *
+    * @return The request path.
+     */
     public String getPath() {
         return path;
     }
 
+    /**
+    * Sets the request path that caused the error.
+    *
+    * @param path The request path.
+     */
     public void setPath(String path) {
         this.path = path;
     }
 
+    /**
+    * Gets the error message, typically used for localization.
+    *
+    * @return The error message.
+     */
     public String getMessage() {
         return message;
     }
 
+    /**
+    * Sets the error message.
+    *
+    * @param message The error message.
+     */
     public void setMessage(String message) {
         this.message = message;
     }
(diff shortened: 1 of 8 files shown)


zhx828:
@jfkonecn already looks better!  What do you think between returning text vs returning a json like what we have in oncokb-curation
```
{
    "type": "https://www.oncokb.org/problem/problem-with-message",
    "title": "Not Found",
    "status": 404,
    "detail": "404 NOT_FOUND",
    "path": "/api/articles/13620",
    "message": "error.http.404"
}
```

Could you also check whether the change would cause the client to be updated? You can update in oncokb-public using `updateOncoKbAPI`. You have to update the url though to point to your local oncokb core instance.

jfkonecn:
> @jfkonecn already looks better! What do you think between returning text vs returning a json like what we have in oncokb-curation
> 
> ```
> {
>     "type": "https://www.oncokb.org/problem/problem-with-message",
>     "title": "Not Found",
>     "status": 404,
>     "detail": "404 NOT_FOUND",
>     "path": "/api/articles/13620",
>     "message": "error.http.404"
> }
> ```
> 
> Could you also check whether the change would cause the client to be updated? You can update in oncokb-public using `updateOncoKbAPI`. You have to update the url though to point to your local oncokb core instance.

It looks like this:
```json
{
    "title": "400 : [{\"type\":\"https://www.oncokb.org/problem/problem-with-message\",\"title\":\"400\",\"status\":400,\"detail\":\"400 BAD_REQUEST hgvsg is invalid.\",\"path\":\"/oncokb/api/v1/annotate/mutations/byHGVSg\",\"message\":\"error.http.400\"}]",
    "status": 400,
    "message": "error.",
    "params": ""
}
```

jfkonecn:
@zhx828  this is ready for you to review.

web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ApiHttpError.java
```diff
@@ -0,0 +1,71 @@
+package org.mskcc.cbio.oncokb.controller.advice;
+
+import org.springframework.http.HttpStatus;
+
+public class ApiHttpError {
+    private String type;
```
zhx828, line 6: Type needs to be removed, otherwise it will still show up in the response.

zhx828 (changes requested):
- I tried to throw ` throw new ApiHttpErrorException("hgvsg is missing.", HttpStatus.BAD_REQUEST);`, it returns the following, which is not what I expect to see in the error response.

```
{
  "title" : "400 : [{\"title\":\"400\",\"status\":400,\"detail\":\"hgvsg is missing.\",\"path\":\"/oncokb/api/private/utils/variantAnnotation\",\"message\":\"error.http.400\"}]",
  "status" : 400,
  "message" : "error.",
  "params" : ""
}
```

- what's the rational of adding the new exception handling for some of the private endpoints?

jfkonecn:
> * I tried to throw ` throw new ApiHttpErrorException("hgvsg is missing.", HttpStatus.BAD_REQUEST);`, it returns the following, which is not what I expect to see in the error response.
> 
> ```
> {
>   "title" : "400 : [{\"title\":\"400\",\"status\":400,\"detail\":\"hgvsg is missing.\",\"path\":\"/oncokb/api/private/utils/variantAnnotation\",\"message\":\"error.http.400\"}]",
>   "status" : 400,
>   "message" : "error.",
>   "params" : ""
> }
> ```
> 
> * what's the rational of adding the new exception handling for some of the private endpoints?

@zhx828 is this from the proxy or directly from core?

zhx828:
> > * I tried to throw ` throw new ApiHttpErrorException("hgvsg is missing.", HttpStatus.BAD_REQUEST);`, it returns the following, which is not what I expect to see in the error response.
> > 
> > ```
> > {
> >   "title" : "400 : [{\"title\":\"400\",\"status\":400,\"detail\":\"hgvsg is missing.\",\"path\":\"/oncokb/api/private/utils/variantAnnotation\",\"message\":\"error.http.400\"}]",
> >   "status" : 400,
> >   "message" : "error.",
> >   "params" : ""
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * what's the rational of adding the new exception handling for some of the private endpoints?
> 
> @zhx828 is this from the proxy or directly from core?
> 
> The rational is to keep things consistent, but I'm okay with reverting the change.

This is from the proxy. I think the public should return the same ApiError instead of nesting under the title. Thoughts?

jfkonecn:
> > > * I tried to throw ` throw new ApiHttpErrorException("hgvsg is missing.", HttpStatus.BAD_REQUEST);`, it returns the following, which is not what I expect to see in the error response.
> > > 
> > > ```
> > > {
> > >   "title" : "400 : [{\"title\":\"400\",\"status\":400,\"detail\":\"hgvsg is missing.\",\"path\":\"/oncokb/api/private/utils/variantAnnotation\",\"message\":\"error.http.400\"}]",
> > >   "status" : 400,
> > >   "message" : "error.",
> > >   "params" : ""
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > * what's the rational of adding the new exception handling for some of the private endpoints?
> > 
> > 
> > @zhx828 is this from the proxy or directly from core?
> > The rational is to keep things consistent, but I'm okay with reverting the change.
> 
> This is from the proxy. I think the public should return the same ApiError instead of nesting under the title. Thoughts?

100% agree, but I'm more asking because I'd have to make a change to public.

zhx828:
> > > > * I tried to throw ` throw new ApiHttpErrorException("hgvsg is missing.", HttpStatus.BAD_REQUEST);`, it returns the following, which is not what I expect to see in the error response.
> > > > 
> > > > ```
> > > > {
> > > >   "title" : "400 : [{\"title\":\"400\",\"status\":400,\"detail\":\"hgvsg is missing.\",\"path\":\"/oncokb/api/private/utils/variantAnnotation\",\"message\":\"error.http.400\"}]",
> > > >   "status" : 400,
> > > >   "message" : "error.",
> > > >   "params" : ""
> > > > }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > 
> > > > * what's the rational of adding the new exception handling for some of the private endpoints?
> > > 
> > > 
> > > @zhx828 is this from the proxy or directly from core?
> > > The rational is to keep things consistent, but I'm okay with reverting the change.
> > 
> > 
> > This is from the proxy. I think the public should return the same ApiError instead of nesting under the title. Thoughts?
> 
> 100% agree, but I'm more asking because I'd have to make a change to public.

Let's do that.

web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ApiHttpError.java
```diff
@@ -0,0 +1,71 @@
...
+    private String title;
+    private int status;
+    private String detail;
+    private String path;
+    private String message;
```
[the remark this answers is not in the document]
jfkonecn: Based on what I've seen is that `message` is a generic error message and `detail` is specifics about why something failed.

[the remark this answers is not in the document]
zhx828: @jfkonecn can we add some comments here explaining the usage? Maybe also give an example?

web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ApiHttpError.java
```diff
@@ -0,0 +1,142 @@
...
+    /** The HTTP status code of the error (e.g., 404, 500). */
+    private int status;
+
+    /** A detailed message explaining the error. */
+    private String detail;
+
+    /** The request path that caused the error. */
+    private String path;
+
+    /** A message key that is standardize to help with client side error handling, such as "error.http.404". */
```
zhx828, line 34: Is this true? This is a message key and is standardized?

jfkonecn: It's my best guess as to why it exist. I could be bad a googling, but I couldn't find what it's supposed to mean. Honestly `detail` is the only field that needs to be there. All the other information is already in a http response.

zhx828 (approved):

PR #3777, opened 2024-08-14
Group relevant alterations based on alteration without exclusion criteria

zhx828:
When alteration has exclusion criteria and however is included in relevantAlterations, that means the alteration in query is not excluded. Therefore the relevant alteration should be treated as the same to the one without exclusion criteria. For instance, Oncogenic Mutations and Oncogenic Mutations {excluding V600K} as relevant alterations when alteration in query is V600E, both relevant alterations are the same.


core/src/main/java/org/mskcc/cbio/oncokb/util/SummaryUtils.java
```diff
@@ -121,20 +121,45 @@ public int compare(Alteration x, Alteration y) {
...
+            Map<String, LinkedHashSet<Alteration>> groupedRelevantAlterations = new HashMap<>();
             for (Alteration alt : relevantAlterations) {
-                tumorTypeSummary = getRelevantTumorTypeSummaryByAlt(evidenceType, alt, matchedTumorType, relevantTumorTypes);
-                if (tumorTypeSummary != null) {
-                    break;
+                String uniqRelevantAlt = AlterationUtils.removeExclusionCriteria(alt.getAlteration());
+                uniqRelevantAlts.add(uniqRelevantAlt);
+                if (!groupedRelevantAlterations.containsKey(uniqRelevantAlt)) {
+                    groupedRelevantAlterations.put(uniqRelevantAlt, new LinkedHashSet<>());
+                }
```
calvinlu3, line 139: Minor - you can use putIfAbsent.

calvinlu3 (approved):

PR #3778, opened 2024-08-14
Do not include special cancer type when finding relevant for evidence

zhx828:
Evidence relevant cancer types are defined based on cancer type that curator selects and it's upward/downward cancer types on OncoTree based on the level.


calvinlu3 (approved):
Thanks!

PR #3780, opened 2024-08-15
Update logic to get tumor type summary

zhx828:
We want to include alteration that is considered the same weight with exact match. We should send all matched alterations to get evidences which then will be sorted based on relevant tumor types order. This guarantees the most relevant tumor type evidence will be used first


calvinlu3 (approved):
LGTM

PR #3788, opened 2024-09-10
Update alteration name conversion logics

zhx828:
- Do not convert any two consecutive uppercases to lowercase for mutation name
- Do not convert word if it's all upper cases

This fixes https://github.com/oncokb/oncokb-pipeline/issues/522


calvinlu3 (approved):
LGTM

PR #3797, opened 2024-10-14
Improve performance of genomicLocationShouldAnnotate

calvinlu3:
Related to https://github.com/oncokb/oncokb-pipeline/issues/553

#### Performance Benchmark
- WES - 13100 variants 31s (422 v/s); 1127 variants were within oncokb coverage (8%) 
- WGS - 1447800 variants: 1362s (1062v/s); 43993 variants were within oncokb coverage (3%) 
image

#### Changes:
1. Create a `Map<String, Set<EnsemblGene>>`, where the key is the chromosome. This allows us to only search the set of EnsemblGenes for one chromosome and allows for short circuiting. There is no reason to check the range if the chromosome doesn’t match the query. 
2. For each reference genome, create two maps. This will reduce the size of Set<EnsemblGene> by ½. 
3. Eliminate the NototationConverter.hgvsgToGenomicLocation() call in oncokb core. It seems that this conversion is only used so that we can grab the chromosome and the start+end to check if there is a range intersect. This can be done without the use of regex.

#### Performance Benchmark
- WES – 13100 variants 25s (524v/s) 
- WGS – 1447800 variants: 1019s (1420v/s) 

image

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
@@ -122,11 +122,7 @@ public class CacheFetcher {
                     continue;
                 }
                 String chromosome = ensemblGene.getChromosome();
-
-                if (!chromosomeEnsemblGenesMap.containsKey(chromosome)) {
-                    chromosomeEnsemblGenesMap.put(chromosome, new HashSet<>());
-                }
-
+                chromosomeEnsemblGenesMap.putIfAbsent(chromosome, new HashSet<>());
                 chromosomeEnsemblGenesMap.get(chromosome).add(ensemblGene);
             }
         }
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
@@ -103,6 +103,27 @@ public class CacheFetcher {
         return oncokbTranscriptService.findTranscriptGenesBySymbols(CacheUtils.getAllGenes().stream().filter(gene -> gene.getEntrezGeneId() > 0).map(gene -> gene.getEntrezGeneId().toString()).collect(Collectors.toList()));
     }
 
+    @Cacheable(cacheResolver = "generalCacheResolver", keyGenerator = "concatKeyGenerator")
+    public Map<String, Set<EnsemblGene>> getCanonicalEnsemblGenesByChromosome(ReferenceGenome referenceGenome) throws ApiException {
+        Map<String, Set<EnsemblGene>> chromosomeEnsemblGenesMap = new HashMap<>();
+        Set<org.oncokb.oncokb_transcript.client.Gene> allTranscriptGenes = getAllTranscriptGenes();
+        for (org.oncokb.oncokb_transcript.client.Gene gene : allTranscriptGenes) {
+            for (EnsemblGene ensemblGene : gene.getEnsemblGenes()) {
+                if (!ensemblGene.getCanonical() || !ensemblGene.getReferenceGenome().equals(referenceGenome.name())) {
+                    continue;
+                }
+                String chromosome = ensemblGene.getChromosome();
+
+                if (!chromosomeEnsemblGenesMap.containsKey(chromosome)) {
+                    chromosomeEnsemblGenesMap.put(chromosome, new HashSet<>());
+                }
+
+                chromosomeEnsemblGenesMap.get(chromosome).add(ensemblGene);
+            }
+        }
+        return chromosomeEnsemblGenesMap;
+    }
+
     private List<CancerGene> getCancerGeneList() throws ApiException, IOException {
         List<CancerGene> cancerGenes = CancerGeneUtils.getCancerGeneList();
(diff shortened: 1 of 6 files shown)


calvinlu3:
@zhx828 With in-memory cache, it takes takes 25s to annotate 13100 variants (524v/s). With redis, it takes 26.89s to annotate 13100 variants (487v/s). About a 7% decrease in performance

calvinlu3:
We will covert to in-memory cache

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -103,6 +108,32 @@ public Set<org.oncokb.oncokb_transcript.client.Gene> getAllTranscriptGenes() thr
...
+        for (org.oncokb.oncokb_transcript.client.Gene gene : allTranscriptGenes) {
+            for (EnsemblGene ensemblGene : gene.getEnsemblGenes()) {
+                if (!ensemblGene.getCanonical() || !ensemblGene.getReferenceGenome().equals(referenceGenome.name())) {
+                    continue;
+                }
+                String chromosome = ensemblGene.getChromosome();
+
+                if (!chromosomeEnsemblGenesMap.containsKey(chromosome)) {
+                    chromosomeEnsemblGenesMap.put(chromosome, new HashSet<>());
+                }
```
zhx828, line 128: putIfAbsent

core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
```diff
@@ -860,4 +861,44 @@ public static Integer compareAnnotationSearchQueryType(AnnotationSearchQueryType
...
+
+    public static int findDigitEndIndex(String str, int startIndex) {
+        int i = startIndex;
+        while (i < str.length() && Character.isDigit(str.charAt(i))) {
+            i++;
+        }
+        return i;
+    }
+
+    public static GenomicLocation parseChromosomeAndRangeFromHGVSg(String hgvsg) {
```
zhx828, line 873: Why are we copying the method over?

calvinlu3: The original implementation used the NotationConverter to parse HGVSg and the regex was the main bottleneck. Also, the `genomicLocationShouldBeAnnotated` only used the chromosome and ranges, so we don't need to parse the entire HGVSg. I believe the parseHGVSg was about 7% of the running state.

calvinlu3: I can revert to using NotationConverter if this is not that big of an issue.

zhx828: Can we add some test cases here?

core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
```diff
@@ -860,4 +861,44 @@ public static Integer compareAnnotationSearchQueryType(AnnotationSearchQueryType
...
+        if (hgvsg == null) {
+            return null;
+        }
+        GenomicLocation location = new GenomicLocation();
+        int start, end;
+
+        // Step 1: Split by ":g."
+        String[] parts = hgvsg.split(":g\\.");
```
zhx828, line 881: Does parts guarantee to have two elements?

core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
```diff
@@ -860,4 +861,44 @@ public static Integer compareAnnotationSearchQueryType(AnnotationSearchQueryType
             return asc ? -1 : 1;
         return (PRIORITIZED_QUERY_TYPES.indexOf(q1) - PRIORITIZED_QUERY_TYPES.indexOf(q2)) * (asc ? 1 : -1);
     }
+
+    public static int findDigitEndIndex(String str, int startIndex) {
```
zhx828, line 865: If we gonna make this public, we need to add more boundary tests.

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java
```diff
@@ -103,6 +108,28 @@ public Set<org.oncokb.oncokb_transcript.client.Gene> getAllTranscriptGenes() thr
         return oncokbTranscriptService.findTranscriptGenesBySymbols(CacheUtils.getAllGenes().stream().filter(gene -> gene.getEntrezGeneId() > 0).map(gene -> gene.getEntrezGeneId().toString()).collect(Collectors.toList()));
     }
 
+    public Map<String, Set<EnsemblGene>> getCanonicalEnsemblGenesByChromosome(ReferenceGenome referenceGenome) throws ApiException {
```
zhx828, line 111: Is this method used at all?

zhx828 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/util/MainUtils.java
calvinlu3: Updated condition to check if both parts are available

zhx828 (approved):
LGTM, thank you!

Issue #3807, opened 2024-10-29
Batching genome nexus requests for byGenomicChange and byHGVSg


zhx828:
Per user's experiment, if we could batch the requests into one, it will gain significant performance improvement. We currently split the POST request to GET for GN annotation.

PR #3809, opened 2024-11-14
Use GN POST to annotate genomic changes

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3807

This change improves genomic change annotation significantly.

### Before
Average in annotating 159 variants takes 5679ms
master-jmeter

### After
Average in annotating 159 variants takes 482ms
batch-gn-jmeter

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
@@ -136,69 +136,88 @@ public class GenomeNexusUtils {
     }
 
     public static TranscriptConsequenceSummary getTranscriptConsequence(GNVariantAnnotationType type, String query, ReferenceGenome referenceGenome) throws ApiException {
-        if (StringUtils.isEmpty(query) || StringUtils.isEmpty(query.replace(",", ""))) {
-            return null;
+        List<VariantAnnotation> annotations = new ArrayList<>();
+        if (type == GNVariantAnnotationType.GENOMIC_LOCATION) {
+            annotations.addAll(getGenomicLocationVariantsAnnotation(Collections.singletonList(GenomeNexusUtils.convertGenomicLocation(query)), referenceGenome));
+        } else if (type == GNVariantAnnotationType.HGVS_G) {
+            annotations.addAll(getHgvsgVariantsAnnotation(Collections.singletonList(query), referenceGenome));
         }
-        VariantAnnotation annotation = getVariantAnnotation(type, query, referenceGenome);
-        return getConsequence(annotation, referenceGenome);
+        Map<String, TranscriptConsequenceSummary> map = getTranscriptsConsequence(annotations, referenceGenome);
+        if (map.isEmpty()) {
+            return null;
+        } else {
+            return map.get(map.keySet().iterator().next());
+        }
     }
 
-    public static GenomeNexusAnnotatedVariantInfo getAnnotatedVariantFromGenomeNexus(GNVariantAnnotationType type, String query, ReferenceGenome referenceGenome) throws ApiException {
-        if (StringUtils.isEmpty(query) || StringUtils.isEmpty(query.replace(",", ""))) {
-            return null;
+    public static Map<String, TranscriptConsequenceSummary> getTranscriptsConsequence(List<VariantAnnotation> annotations, ReferenceGenome referenceGenome) throws ApiException {
+        Map<String, TranscriptConsequenceSummary> result = new HashMap<>();
+        for (VariantAnnotation annotation : annotations) {
+            result.put(annotation.getOriginalVariantQuery(), getConsequence(annotation, referenceGenome));
         }
+        return result;
+    }
 
-        GenomeNexusAnnotatedVariantInfo preAnnotatedVariantInfo = new GenomeNexusAnnotatedVariantInfo();
-        preAnnotatedVariantInfo.setOriginalVariantQuery(query);
-        preAnnotatedVariantInfo.setReferenceGenome(referenceGenome);
+    public static List<GenomeNexusAnnotatedVariantInfo> getAnnotatedVariantsFromGenomeNexus(GNVariantAnnotationType type, List<VariantAnnotation> variantsAnnotation, ReferenceGenome referenceGenome) throws ApiException {
+        List<GenomeNexusAnnotatedVariantInfo> result = new ArrayList<>();
 
-        VariantAnnotation annotation = null;
-        annotation = getVariantAnnotation(type, query, referenceGenome);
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
```diff
@@ -215,18 +227,43 @@ private static String getIsoform(Gene gene, ReferenceGenome referenceGenome) {
...
     }
 
-    private static VariantAnnotation getVariantAnnotation(GNVariantAnnotationType type, String query, ReferenceGenome referenceGenome) throws ApiException {
-        VariantAnnotation variantAnnotation = null;
-        if (query != null && type != null) {
+    public static String chromosomeNormalizer(String chromosome) {
+        return chromosome.trim().replace("chr", "").replace("23", "X").replace("24", "Y");
+    }
+
+    private static GenomicLocation convertGenomicLocation(String genomicLocation) {
```
calvinlu3, line 234: Was this method introduced for performance reasons? NotationConverter also has a method to parse genomic location

core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
```diff
@@ -215,18 +227,43 @@ private static String getIsoform(Gene gene, ReferenceGenome referenceGenome) {
...
+        List<VariantAnnotation> variantsAnnotation = null;
+        if (queries != null && type != null) {
             List<String> gnFields = new ArrayList<>();
             gnFields.add("annotation_summary");
             if (type.equals(GNVariantAnnotationType.HGVS_G)) {
-                variantAnnotation = getAnnotationControllerApi(referenceGenome).fetchVariantAnnotationGET(query, MSK_ISOFORM_OVERRIDE, null, gnFields);
+                variantsAnnotation = getAnnotationControllerApi(referenceGenome).fetchVariantAnnotationPOST(queries, MSK_ISOFORM_OVERRIDE, null, gnFields);
             } else {
-                variantAnnotation = getAnnotationControllerApi(referenceGenome).fetchVariantAnnotationByGenomicLocationGET(query, MSK_ISOFORM_OVERRIDE, null, gnFields);
+                variantsAnnotation = getAnnotationControllerApi(referenceGenome).fetchVariantAnnotationByGenomicLocationPOST(queries.stream().map(GenomeNexusUtils::convertGenomicLocation).collect(Collectors.toList()), MSK_ISOFORM_OVERRIDE, null, gnFields);
```
calvinlu3, line 263: If `convertGenomicLocation` returns null, is that supported by the GN POST request? I used the GN swagger to make a POST with an array `[{valid GL object}, null]` and I get a 404 error. We should filter out the null entries?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -557,23 +612,33 @@ private IndicatorQueryResp getIndicatorQueryFromGenomicLocation(
...
             evidenceTypes,
-                false
+            false
         );
     }
 
+    private Map<String, Alteration> getAlterationsFromGenomeNexus(GNVariantAnnotationType type, ReferenceGenome referenceGenome, List<String> queries) throws ApiException, org.genome_nexus.ApiException {
+        List<String> queriesToGN = new ArrayList<>();
+
+        for (String query : queries) {
```
calvinlu3, line 622: Should we filter out queries that are empty strings at this step. This looks like the entry point to get GN annotation.

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateGenomeNexusController.java
```diff
@@ -74,16 +75,20 @@ public ResponseEntity<TranscriptResult> getTranscript(
...
 
         if (body == null) {
             throw new ApiHttpErrorException("The request body is missing.", HttpStatus.BAD_REQUEST);
         } else {
-            for (AnnotateMutationByHGVSgQuery query : body) {
-                GenomeNexusAnnotatedVariantInfo resp = GenomeNexusUtils.getAnnotatedVariantFromGenomeNexus(GNVariantAnnotationType.HGVS_G, query.getHgvsg(), query.getReferenceGenome());
-                result.add(resp);
+            ReferenceGenome referenceGenome = null;
+            if (!body.isEmpty()) {
+                referenceGenome = body.get(0).getReferenceGenome();
```
calvinlu3, line 86: Its not guaranteed that the reference genome is the same for the entire List<AnnotateMutationByHGVSgQuery>, but we're using the first ReferenceGenome here.

If we were to support a POST query with potentially different ReferenceGenome, then we probably have to partition the list into two, get GN annotation, then merge together?

core/src/test/java/org/mskcc/cbio/oncokb/util/IndicatorUtilsTest.java
calvinlu3, line 1: Potentially some additional tests cases for GNVariantAnnotationType genomic location and some edge cases would make this test suite more complete.

calvinlu3 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
zhx828: Issue seems to be util method can't use the NotationConverter class.

core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
zhx828: My thinking is we should not. We've been trying to persist the query location in the list. WI'm thinking to return null for the null input.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
zhx828: `genomicLocationShouldBeAnnotated` will filter out the empty query.

core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
calvinlu3: Gotcha, persisting the index is valid. However, do you know if passing in a null value in the POST body array will throw an exception for the entire request even if there are valid queries to annotate?

zhx828: null in request body or in response?

calvinlu3 (approved):
LGTM

PR #3823, opened 2024-12-16
Add error handlers to provide proper API code

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/612

Instead of 500 internal server error, we should return the API code from GN.


web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ExceptionTranslator.java
```diff
@@ -0,0 +1,48 @@
...
+        return new ResponseEntity<>(error, ex.getHttpStatus());
+    }
+
+    @ExceptionHandler(org.oncokb.oncokb_transcript.ApiException.class)
+    public final ResponseEntity<ApiHttpError> handleTranscriptApiException(org.oncokb.oncokb_transcript.ApiException ex, WebRequest request) {
+        String path = request.getDescription(false).replaceFirst("uri=", "");
+        HttpStatus httpStatus = HttpStatus.INTERNAL_SERVER_ERROR;
+        try {
+            httpStatus = HttpStatus.valueOf(ex.getCode());
+        } catch (IllegalArgumentException e) {}
```
jfkonecn, line 29: Do we have the ability to log this case? Would it expected to enter this catch block?

web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ExceptionTranslator.java
```diff
@@ -0,0 +1,48 @@
...
+    public final ResponseEntity<ApiHttpError> handleTranscriptApiException(org.genome_nexus.ApiException ex, WebRequest request) {
+        String path = request.getDescription(false).replaceFirst("uri=", "");
+        HttpStatus httpStatus = HttpStatus.INTERNAL_SERVER_ERROR;
+        String message = ex.getMessage();
+        if (StringUtils.isEmpty(message)) {
+            message = "Error annotating variant(s) using Genome Nexus";
+        }
+        try {
+            httpStatus = HttpStatus.valueOf(ex.getCode());
+        } catch (IllegalArgumentException e) {}
```
jfkonecn, line 44: same as above

web/src/main/java/org/mskcc/cbio/oncokb/controller/advice/ExceptionTranslator.java
calvinlu3: It will enter this catch block when `ex.getCode() == 0`. 0 is the default status code for ApiException classes and not a valid http status code, so the default 500 error is thrown instead.

calvinlu3: Logging is definitely possible, but sometimes the http code is optional. Maybe a comment here explaining the decision would suffice? What do you think @jfkonecn?

jfkonecn: yeah I'm good with comment. An empty catch block always makes me think something is wrong.

jfkonecn (approved):

PR #3825, opened 2024-12-17
classify complex delins mutations with stop codon inserted as truncation

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/629

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -300,7 +300,7 @@ public final class AlterationUtils {
 
         proteinChange = proteinChange.trim();
 
-        Pattern p = Pattern.compile("([A-Z]?)([0-9]+)(_[A-Z]?([0-9]+))?(delins|ins|del)([A-Z0-9]*)", Pattern.CASE_INSENSITIVE);
+        Pattern p = Pattern.compile("([A-Z]?)([0-9]+)(_[A-Z]?([0-9]+))?(delins|ins|del)([A-Z0-9\\*]*)", Pattern.CASE_INSENSITIVE);
         Matcher m = p.matcher(proteinChange);
             if (m.matches()) {
                 if (m.group(1) != null && m.group(3) == null) {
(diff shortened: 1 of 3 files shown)


calvinlu3 (changes requested):
I would also add a test in `FindRelevantAlterationsTest.java` to see if truncating mutation is found as the relevant alt. See example on L78.

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -300,7 +300,7 @@ public static void annotateAlteration(Alteration alteration, String proteinChang
 
         proteinChange = proteinChange.trim();
 
-        Pattern p = Pattern.compile("([A-Z]?)([0-9]+)(_[A-Z]?([0-9]+))?(delins|ins|del)([A-Z0-9]*)", Pattern.CASE_INSENSITIVE);
+        Pattern p = Pattern.compile("([A-Z]?)([0-9]+)(_[A-Z]?([0-9]+))?(delins|ins|del)([A-Z0-9\\*]*)", Pattern.CASE_INSENSITIVE);
```
calvinlu3, line 303: Adding the stop codon (*) in group 6 will match patterns that have multiple asterisks, which is not valid.

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -324,7 +324,10 @@ public static void annotateAlteration(Alteration alteration, String proteinChang
                     String groupSix = m.group(6);
                     String groupSixWithoutDigits = MainUtils.removeDigits(groupSix);
 
-                    if (groupSixWithoutDigits.length() != groupSix.length() && groupSixWithoutDigits.length() > 0) {
+                    if (groupSixWithoutDigits.contains("*")) {
```
calvinlu3, line 327: If we adjust the regex in the comment on L303, then we can continue using contains. Otherwise currently, it would make more sense to use endsWith.

bprize15: Answered in standup

calvinlu3:
> I would also add a test in `FindRelevantAlterationsTest.java` to see if truncating mutation is found as the relevant alt. See example on L78.

This is a slightly different test case that also makes sure the downstream logic are correct. The test you added in AlterationUtilsTest is valid for determining if the right consequence is added.

calvinlu3 (approved):
LGTM! Just one small test request

PR #3830, opened 2025-01-15
No frontend

jfkonecn:


calvinlu3:
@jfkonecn Do you remember what the conflict was?

jfkonecn:
> @jfkonecn Do you remember what the conflict was?

It seemed like Ben's changes were the issue, but they were very subtle.
It flagged a lot of things that seemed to be the same.
If you want we can have a call about it.
I can make a new branch and we can see where the changes are.

calvinlu3:
> > @jfkonecn Do you remember what the conflict was?
> 
> It seemed like Ben's changes were the issue, but they were very subtle. It flagged a lot of things that seemed to be the same. If you want we can have a call about it. I can make a new branch and we can see where the changes are.

We can talk after standup

calvinlu3 (approved):

PR #3833, opened 2025-01-27
Added Genomic Indicator Endpoint

jfkonecn:


web/src/main/java/org/mskcc/cbio/oncokb/controller/GeneController.java
```diff
@@ -53,6 +55,48 @@ List<Gene> getGene(
...
+        , @RequestParam(value = "hugoSymbol", required = false) List<String> hugoSymbols
+        , @RequestParam(value = "fields", required = false) String fields
+    ) {
+        if (entrezGeneIds == null && hugoSymbols == null) {
+            return new ArrayList<>();
+        }
+
+        Set<Gene> genes = new HashSet<>();
+        if (entrezGeneIds != null) {
+            for (Integer enterz : entrezGeneIds) {
```
calvinlu3, line 71:
```suggestion
            for (Integer entrez : entrezGeneIds) {
```

web/src/main/java/org/mskcc/cbio/oncokb/controller/GeneController.java
```diff
@@ -53,6 +55,48 @@ List<Gene> getGene(
...
+
+        for (Map.Entry<Gene, Set<Evidence>> entry : map.entrySet()) {
+            for (Evidence evidence : entry.getValue()) {
+                if (evidence.getEvidenceType() == EvidenceType.GENOMIC_INDICATOR) {
+                    evidenceList.add(evidence);
+                }
+            }
+        }
+
+        return JsonResultFactory.getEvidence(evidenceList, fields);
```
calvinlu3, line 97: I would just only return `id, gene, name, description` fields and remove the `fields` parameter. Seems like too much info for them to see all evidence fields.

calvinlu3 (changes requested):

calvinlu3 (approved):

PR #3841, opened 2025-02-04
Fix transcript selection logic and return canonical transcript id in API response

calvinlu3:
Resolves https://github.com/oncokb/oncokb-pipeline/issues/642

### Changes
- Update logic for selecting the transcriptConsequenceSummary that matches the oncokb canonical transcript
- Return transcript id via API to clearly denote which transcript was used
```
"query": {
    "id": null,
    "referenceGenome": "GRCh37",
    "hugoSymbol": "BRAF",
    "entrezGeneId": 673,
    "alteration": "V600E",
    "alterationType": null,
    "svType": null,
    "tumorType": null,
    "consequence": "missense_variant",
    "proteinStart": 600,
    "proteinEnd": 600,
    "hgvs": "7:g.140453136A>T",
    "hgvsInfo": null,
    "canonicalTranscript": "ENST00000288602"
  }
  ```
- Return an error message to inform user that queried variant does not fall into any oncokb canonical transcript range
```
"query": {
    "id": null,
    "referenceGenome": "GRCh37",
    "hugoSymbol": null,
    "entrezGeneId": null,
    "alteration": "",
    "alterationType": null,
    "svType": null,
    "tumorType": null,
    "consequence": null,
    "proteinStart": null,
    "proteinEnd": null,
    "hgvs": "7:g.101923315_101923316insC",
    "hgvsInfo": "This variant does not occur within a canonical transcript range of an OncoKB cancer gene.",
    "canonicalTranscript": null
  }
  ```

### Before Review
- [x] Use the WGS/WES studies to compare the api responses to make sure annotations have not changed.


jfkonecn (approved):

bprize15 (approved):
LGTM

calvinlu3:
@jjgao Before I merge the PR - was wondering if the API response satisfied your comment from the ticket
> I agree with your solution, but wonder if there is any place that we can add a little comment such as "the variant is outside of the range of the selected transcript ENSTxxxx"
Particularly the message I added: `This variant does not occur within a canonical transcript range of an OncoKB cancer gene.`

jjgao:
Thanks, @calvinlu3! 

Maybe simplify it to: `This variant does not occur within a gene or transcript annotated in OncoKB.`

calvinlu3:
@bprize15 Re-requesting review because I had to resolve some conflicts with your sample api changes.
Changes:
1. I reverted your change `hgvsg -> hgvs` back because this is considered a breaking API change. If we release the changes, then people annotating using our public endpoints will not get annotations back because they are still using "hgvsg" in their POST body vs "hgvs". Also, I think for the endpoint we need to continue using `hgvsg` because that is a specific format (7:g.123...). `hgvs` to me refers to all the formats.
2. Can you review `AnnotationsApiController.java` because you move the logic into methods like `annotateMutationsByHGVSg` and `annotateMutationsByGenomicLocation` for reusability. My PR changes the the use of `Alteration -> TranscriptSummaryAlterationResult`, so I copied my revised implementation over to your methods.
3. Can you also review this change (https://github.com/oncokb/oncokb/pull/3841/commits/800720eeb36b75aec7cf259aa54e4eec59b305d5). It is what we discussed during the weekly meeting.

Sorry for the review again!

core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
```diff
@@ -324,48 +329,60 @@ private static TranscriptConsequenceSummary getConsequence(VariantAnnotation var
...
-            summary = summaries.iterator().next();
-        } else if (summaries.size() > 1) {
-            summary = pickTranscriptConsequenceSummary(summaries, variantAnnotation.getAnnotationSummary().getCanonicalTranscriptId(), variantAnnotation.getMostSevereConsequence());
+        TranscriptSummaryAlterationResult selectedAnnotationResult;
+        if (annotationResult.size() == 1) {
+            // If there is only one matching transcript, use the summary.
+            selectedAnnotationResult = annotationResult.iterator().next();
+        } else if (annotationResult.size() > 1) {
+            // Prioritize based on most_severe_consequence
+            annotationResult = filterTranscriptConsequenceSummaryByMostSevereConsequence(annotationResult, variantAnnotation.getMostSevereConsequence());
```
bprize15, line 345: I remember Xiang mentioning something about choosing Ensembl canonical after checking by most severe. I'm fine with either way, but wondering if you have a reason for or against this.

bprize15 (approved):
One question, but LGTM

core/src/main/java/org/mskcc/cbio/oncokb/util/GenomeNexusUtils.java
calvinlu3: Talked over the logic with her yesterday and she said that choosing Ensembl canonical was mentioned because that's how GN transcript selection works. I clarified that we filter OncoKB canonical transcripts, so unless the Ensembl canonical is also the OncoKB canonical, we will not choose the Ensembl canonical. She was more suggesting the scenario where OncoKB canonical transcript summary does not contain the most severe consequence. In that case, she suggested to prioritize based on variant consequence severity.

Issue #3827, opened 2024-12-18
Ensembl versioning


calvinlu3:
### Tasks
- [ ] Setup a local VEP on a static Ensembl version
- [x] Ensure GN Mongo cache only contains annotations from a single Ensembl version
- [x] Allow oncokb endpoint to take in an `ensembl_version` parameter (Not planned)

bprize15:
- Add `/info/software` to local VEP to mimic Ensembl API
- Genome Nexus `/version` will hit `/info/software` to populate VEP version
- Speak to Avery about consolidating "cache" and "server" versions into one
- Speak to Avery about the following options for populating "static" flag
    - Delete "static"
    - Return "static" from VEP
    - Move "static" to `application.properties`

PR #3863, opened 2025-02-27
add genome nexus info to info endpoint

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3827


core/src/main/java/org/mskcc/cbio/oncokb/model/OncoKBInfo.java
```diff
@@ -49,6 +52,12 @@ public OncoKBInfo() {
         } else {
             this.publicInstance = false;
         }
+
+        try {
+            this.genomeNexus = GenomeNexusUtils.getParsedGenomeNexusVersion();
+        } catch (RestClientException e) {
+            this.genomeNexus = null;
```
jfkonecn, line 59: @calvinlu3 do you think it'll be confusing to set genomeNexus to null on a fail API request? I feel like throwing a 5xx error isn't really what we want so this should be fine.

jfkonecn (approved):

calvinlu3 (approved):

Issue #3869, opened 2025-03-03
Add health endpoint to validate in memory cache state


calvinlu3:
We need to add a health check endpoint to ensure that our in-memory caches are valid.

### Tasks
- Implement a `/health/cache` endpoint 
- Endpoint should check the size of the in memory data structures with the expected size

PR #3871, opened 2025-03-04
Update startup logs and add health endpoint

calvinlu3:
Resolves https://github.com/oncokb/oncokb/issues/3869

### Changes
Add a health endpoint that fetches the data from MySQL again and compares it with the in-memory caches populated during app startup.

```
## Output from logs when health endpoint fails

Genes cache: Expected 906, but is 914
Alterations cache: Expected 19260, but is 19277
Drugs cache: Expected 148, but is 149
Gene-based Evidences cache: Expected 906, but is 914
```

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
@@ -263,7 +263,7 @@ public class UtilsApiController implements UtilsApi {
         return new ResponseEntity<>(sb.toString(), HttpStatus.OK);
     }
 
-    private List<ActionableGene> getAllActionableVariants(Boolean isTextFile) {
+    public List<ActionableGene> getAllActionableVariants(Boolean isTextFile) {
         List<ActionableGene> actionableGeneList = new ArrayList<>();
         Set<Gene> genes = CacheUtils.getAllGenes();
         Map<Gene, Set<ClinicalVariant>> map = new HashMap<>();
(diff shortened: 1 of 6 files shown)


web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -40,6 +40,7 @@
         <url-pattern>/api/v1/*</url-pattern>
         <url-pattern>/api/private/*</url-pattern>
         <url-pattern>/legacy-api</url-pattern>
+        <url-pattern>/health</url-pattern>
```
jfkonecn, line 43: since we'll have a client depend on this I think we should include a version number. Maybe something like `/api/internal/v1/health`?

jfkonecn (approved):

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -263,7 +263,7 @@ public ResponseEntity<String> utilsAllActionableVariantsTxtGet(
         return new ResponseEntity<>(sb.toString(), HttpStatus.OK);
     }
 
-    private List<ActionableGene> getAllActionableVariants(Boolean isTextFile) {
+    public List<ActionableGene> getAllActionableVariants(Boolean isTextFile) {
```
jfkonecn, line 266: Is there a way to refactor this into a utility method? I think having other controllers call each other will be a little confusing since people won't be expecting it.

jfkonecn (changes requested):

jfkonecn (approved):

Issue #3872, opened 2025-03-04
Clear Redis cache after application startup to prevent stale data during Kubernetes deployments


calvinlu3:
### Issue
Currently, our application clears the Redis cache at startup. However, in Kubernetes deployments, old pods remain running until the new pods are fully available. This creates a scenario where:
1. Old pods can still process requests and write stale data back into Redis.
2. New pods may fetch and cache outdated data before all old pods are terminated.

### Tasks
- Move Redis cache clearing logic to post-startup 
- Make sure that during startup in-memory caching (CacheUtils.java), Redis is not utilized. We do not want stale data being read or stored into Redis.

PR #3878, opened 2025-03-05
Add unique Redis key prefix

calvinlu3:
Resolves https://github.com/oncokb/oncokb/issues/3872

Initially, I was thinking to do a post-startup cache clear, but I noticed that there may exists a brief moment in time where old pods and new pods co-exist (depending on timing of readiness probe, and old pod termination in k8s).

Instead, we can add a unique cache prefix (ie `oncokb-core-app:v3.26.0-data:v4.25.0`). We add versioning to the Redis values and avoid cross contamination between old/new pods.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheConfiguration.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheConfiguration.java
@@ -2,6 +2,8 @@ package org.mskcc.cbio.oncokb.cache;
 
 import org.apache.commons.lang3.StringUtils;
 import org.mskcc.cbio.oncokb.cache.keygenerator.ConcatGenerator;
+import org.mskcc.cbio.oncokb.model.OncoKBInfo;
+import org.mskcc.cbio.oncokb.util.CacheUtils;
 import org.mskcc.cbio.oncokb.util.PropertiesUtils;
 import org.mskcc.oncokb.meta.enumeration.RedisType;
 import org.redisson.Redisson;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheConfiguration.java
```diff
@@ -102,12 +104,22 @@ private void setRedisConnectionPoolSize(BaseMasterSlaveServersConfig baseMasterS
...
-            baseConfig.setClientName(appName);
-        } else {
-            baseConfig.setClientName("oncokb-core");
-        }
+        String appNameProperty = PropertiesUtils.getProperties("app.name");
+
+        StringBuilder appName = new StringBuilder();
+        appName.append(StringUtils.isNotEmpty(appNameProperty) ? appNameProperty : "oncokb-core");
+
+        OncoKBInfo oncoKBInfo = new OncoKBInfo();
```
bprize15, line 112: Do we need to create a new OncoKBInfo? This will cause a fetch to genomeNexus. Can we just get the appVersion and dataVersion directly like in OncoKBInfo?

calvinlu3: Good point, let me look into this tomorrow.

calvinlu3: I realized the logic was in the wrong place - moved to CacheNameResolver instead, which is where the redis key is generated.

Also agree with you comment that we do not want to fetch from GN every time the cache name is generated. Addressed that.


Cache key will look something like `"oncokb-core-app:v3.28.0-data:v4.26:general:processQuery:GRCh37:673:BRAF:V600E:null:null:missense_variant:600:600:null:7,140453136,140453136,A,T:null:false:[]:false"`

core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheConfiguration.java
```diff
@@ -2,6 +2,8 @@
 
 import org.apache.commons.lang3.StringUtils;
 import org.mskcc.cbio.oncokb.cache.keygenerator.ConcatGenerator;
+import org.mskcc.cbio.oncokb.model.OncoKBInfo;
```
bprize15, line 5: Don't need these anymore

bprize15 (approved):
Just some unused imports, otherwise LGTM

Issue #3794, opened 2024-10-02
Return mutation summary if available instead of the default


zhx828:
Use case of mutation summary override: It allows us to alter the mutation summary for resistance variants. 
This curated summary will replace the programmatically generated mutation summary from `SummaryUtils.variantSummary(...)`

PR #3889, opened 2025-03-27
Use curated mutation summary instead of generated if available

calvinlu3:
Fixes https://github.com/oncokb/oncokb/issues/3794

### Todo:
- [x] How should generated summary in T315G if we override the default summary for T315I?
image
image
   - We decided to continue using generated summary for T315G for the time being

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/SummaryUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/SummaryUtils.java
@@ -257,7 +257,7 @@ public class SummaryUtils {
             return ONCOGENIC_MUTATIONS_DEFAULT_SUMMARY;
         }
         EvidenceBo evidenceBo = ApplicationContextSingleton.getEvidenceBo();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/SummaryUtils.java
```diff
@@ -256,6 +256,15 @@ public static String variantSummary(Gene gene, Alteration exactMatchAlteration,
         if (!StringUtils.isEmpty(query.getAlteration()) && query.getAlteration().toLowerCase().startsWith(InferredMutation.ONCOGENIC_MUTATIONS.getVariant().toLowerCase())) {
             return ONCOGENIC_MUTATIONS_DEFAULT_SUMMARY;
         }
+        EvidenceBo evidenceBo = ApplicationContextSingleton.getEvidenceBo();
+        // Use the curated mutation summary if exists to override the generated summary only for the excact mutation. Subject to change
```
jfkonecn, line 260: excact => exact

jfkonecn:
@calvinlu3 once you fix the typo in the comment and you can merge. I don't need to review it again.

PR #3897, opened 2025-04-11
Added Dev only valdation error message for transcript errors

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/722


web/src/main/java/org/mskcc/cbio/oncokb/api/websocket/CurationValidationApiController.java
```diff
@@ -25,6 +25,8 @@
 
 @ServerEndpoint(value = "/api/websocket/curation/validation")
 public class CurationValidationApiController {
+    private static final String DEV_TEAM_ISSUE = " This is an issue the dev team needs to fix during the data release.";
```
calvinlu3, line 28:
```suggestion
    private static final String DEV_TEAM_ISSUE = "This is an issue the dev team needs to fix during the data release.";
```

calvinlu3 (approved):

PR #3904, opened 2025-04-24
update genome nexus and gn vep versions in docker compose

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3903

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,5 @@
+# NOTE: Image versions in this file were last updated on 4/24/25
+
 version: "3"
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
calvinlu3, line 1: Can you make a note somewhere that the image versions in this docker-compose is current as of some date? Or note that it may not always be the same as our production images.

A user spent a bit of time debugging because they thought images were all latest.

calvinlu3 (approved):

calvinlu3 (approved):

PR #3910, opened 2025-05-12
allow search bar to recognize alterations on different transcripts

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/745

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -1,6 +1,33 @@
 package org.mskcc.cbio.oncokb.util;
 
-import org.apache.commons.lang3.ArrayUtils;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -694,45 +726,49 @@ public static Alteration getAlteration(String hugoSymbol, String alteration, Alt
 
     private static Alteration convertTranscriptConsequenceSummaryToAlteration(TranscriptSummaryAlterationResult transcriptSummaryAlterationResult) {
```
calvinlu3, line 727:
```suggestion
    private static Alteration convertTranscriptSummaryAlterationResultToAlteration(TranscriptSummaryAlterationResult transcriptSummaryAlterationResult) {
```

calvinlu3 (approved):
Looks good!
Just a minor method name change. I didn't name the method properly when I created it.

PR #3911, opened 2025-05-13
add exon to annotation response

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/746


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -822,4 +819,15 @@ private List<IndicatorQueryResp> annotateMutationsByHGVSg(List<AnnotateMutationB
         }
         return result;
     }
+
+    private void addTranscriptAndExonToResponse(IndicatorQueryResp response, TranscriptConsequenceSummary summary) {
+        if (summary != null && StringUtils.isNotEmpty(summary.getTranscriptId())) {
+            if (StringUtils.isNotEmpty(summary.getTranscriptId())) {
```
calvinlu3, line 825: Already checked if summary.getTranscriptId() is null on L824, can remove this.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -822,4 +819,15 @@ private List<IndicatorQueryResp> annotateMutationsByHGVSg(List<AnnotateMutationB
...
+                response.getQuery().setCanonicalTranscript(summary.getTranscriptId());
+            }
+            if (StringUtils.isNotEmpty(summary.getExon())) {
+                response.setExon(summary.getExon().replaceAll("/.*", ""));
```
calvinlu3, line 829: Another option is to use [StringUtils.substringBefore](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringBefore(java.lang.String,int)) (if that is what your regex is doing?). We did a JProfile of core before and saw that regex less performant than looping through the string.

calvinlu3 (approved):
LGTM!
Left two small comments.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
bprize15: Ah, good catch

PR #3914, opened 2025-05-15
Fix TumorType swagger model

calvinlu3:
Related to https://github.com/oncokb/oncokb-transcript/pull/523

We have `org.mskcc.cbio.oncokb.model.TumorType` and `org.mskcc.cbio.oncokb.apiModel.TumorType` annotated with `@ApiModel` for swagger docs to pick up. They have the same name, so one of them will be overriden by the other, so I added "value" property to annotation to distinguish them.


bprize15 (approved):
LGTM

PR #3933, opened 2025-07-17
support hgvsg in database for hgvsg endpoint

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/804

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
@@ -506,38 +506,6 @@ public class AnnotationsApiController {
         return indicatorQueryResp;
     }
 
-    private List<TranscriptSummaryAlterationResult> getAlterationsFromGenomeNexusByGenomicLocation(ReferenceGenome referenceGenome, List<GenomicLocation> queries) throws ApiException, org.genome_nexus.ApiException {
-        List<GenomicLocation> queriesToGN = new ArrayList<>();
-        Map<GenomicLocation, Integer> queryIndexMap = new HashMap<>();
-        for (GenomicLocation query : queries) {
-            if (this.cacheFetcher.genomicLocationShouldBeAnnotated(query, referenceGenome)) {
-                if (!queryIndexMap.containsKey(query)) {
-                    queryIndexMap.put(query, queriesToGN.size());
-                    queriesToGN.add(query);
-                }
-            }
-        }
-        List<org.genome_nexus.client.VariantAnnotation> variantAnnotations = GenomeNexusUtils.getGenomicLocationVariantsAnnotation(queriesToGN, referenceGenome);
-        if(variantAnnotations.size() != queriesToGN.size()){
-            throw new ApiException("Number of variants that have been annotated by GenomeNexus is not equal to the number of queries");
-        }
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
```diff
@@ -818,30 +858,98 @@ private List<IndicatorQueryResp> annotateMutationsByHGVSg(List<AnnotateMutationB
...
-            );
-            resp.getQuery().setId(query.getId());
-            resp.getQuery().setHgvsInfo(alterationInfo.getMessage());
-            result.add(resp);
+    private List<IndicatorQueryResp> annotateMutationsByHGVSg(ReferenceGenome referenceGenome, List<AnnotateMutationByHGVSgQuery> queries) throws ApiException, org.genome_nexus.ApiException {
+        List<String> queriesToGN = new ArrayList<>();
+        Map<String, Integer> queryIndexMap = new HashMap<>();
+        for (AnnotateMutationByHGVSgQuery query : queries) {
+            String hgvsg = query.getHgvsg();
+            if (this.cacheFetcher.hgvsgShouldBeAnnotated(hgvsg, referenceGenome) || AlterationUtils.isValidHgvsp(hgvsg)) {
```
calvinlu3, line 883: Wondering if there is a typo here and should be `isValidHGVsg`?

calvinlu3 (changes requested):
- When querying `7:g.140453136A>T`, the variantSummary field uses a lowercase `a` instead of capital
- Should we still set the `query.canonicalTranscript` field even if g. is explicitly curated
- Thinking we should populate the `query.hgvs` field even if g. is not explicitly curated to match behavior regular g. behavior.
- This is the least of my concern: there is some duplicate logic in `annotateMutationsByGenomicChange(ReferenceGenome referenceGenome, List<AnnotateMutationByGenomicChangeQuery> queries)`  and `annotateMutationsByHGVSg(ReferenceGenome referenceGenome, List<AnnotateMutationByHGVSgQuery> queries)`. If there is an easy way to to not repeat code, otherwise please ignore.

bprize15:
> * When querying `7:g.140453136A>T`, the variantSummary field uses a lowercase `a` instead of capital
> * Should we still set the `query.canonicalTranscript` field even if g. is explicitly curated
> * Thinking we should populate the `query.hgvs` field even if g. is not explicitly curated to match behavior of regular g. behavior.
> * This is the least of my concern: there is some duplicate logic in `annotateMutationsByGenomicChange(ReferenceGenome referenceGenome, List<AnnotateMutationByGenomicChangeQuery> queries)`  and `annotateMutationsByHGVSg(ReferenceGenome referenceGenome, List<AnnotateMutationByHGVSgQuery> queries)`. If there is an easy way to to not repeat code, otherwise please ignore.

Will leave transcript out as we discussed. Also, the two methods have subtle differences that make creating an abstraction difficult, so I think we can leave for now, and if we see more repetition in the future, we can refactor.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/AnnotationsApiController.java
bprize15: The `isValidHgvsp` check was there previously. This was when I planned on using this method to get the exon for the protein change. However, since we are no longer doing this (due to performance limitations), I will remove it.

PR #3945, opened 2025-08-05
Allow mvn tests to be ran in docker containers to avoid MySQL database changes

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/841

### Issue:
When running data release, we often run `mvn test` which insert some dummy data (does not cleanup), and that ends up in production.

### Fix
Add option to run mvn test in a docker container with MySQL initialized from a mysql database dump file. Dummy data is not persisted onto actual database that may be used for downstream workflow, like data release.
  - Easier to get mvn test running in a pipeline because you don't need to setup MySQL, docker-compose will spin neccessary services.
  
I was not able to implement rollbacks on database using Transactions and Rollback annotations due to our old spring version. Open to hearing if anyone has ideas on how to cleanup test data cleanly, if we feel strongly about running `mvn test` locally rather than in docker container.

### DIFF
--- a/unit_test/scripts/docker-compose.yml
+++ b/unit_test/scripts/docker-compose.yml
@@ -1,0 +1,36 @@
+services:
+  mysql:
+    image: mysql:8.0.36
+    environment:
+      MYSQL_DATABASE: "oncokb_core_test"
+      MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD}
+      MYSQL_DEFAULT_AUTHENTICATION_PLUGIN: mysql_native_password
+    command: --default-authentication-plugin=mysql_native_password
+    volumes:
+      - type: bind
+        source: ../mysql_dumps
+        target: /docker-entrypoint-initdb.d
+    healthcheck:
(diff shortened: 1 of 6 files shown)


unit_test/scripts/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+services:
+  mysql:
+    image: mysql:8.0.36
+    environment:
+      MYSQL_DATABASE: "oncokb_core_test"
+      MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD}
+      MYSQL_DEFAULT_AUTHENTICATION_PLUGIN: mysql_native_password
+    # Workaround to resolve error (https://stackoverflow.com/questions/50424900/error-client-does-not-support-authentication-protocol-requested-by-server-cons)
+    # TLDR; we need to upgrade our MySQL connect from v5 to v8, so that everything works properly with mysql v8+
+    command: --default-authentication-plugin=mysql_native_password
```
jfkonecn, line 10: @ReshmaRamaiah10 our favorite problem

PR #3955, opened 2025-09-02
Update unit-test yaml

calvinlu3:
### Changes
- Use `shogo82148/actions-setup-mysql@v1` because it allows us to have greater control over MySQL server configuration
- Use `default-authentication-plugin=mysql_native_password` since MySQL connector is still on v5
- Run unit tests on PRs

Ideally we should update the MySQL connector to v8

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -4,7 +4,8 @@ name: Unit Tests
 
 on:
   push:
-    branches: [ master, refactor , next-minor-release ]
+    branches: [master, refactor, next-minor-release]
+  pull_request:
 
 jobs:
   main:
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-test.yml
```diff
@@ -16,11 +17,12 @@ jobs:
         run: sudo service mysql stop # Shutdown the Default MySQL, "sudo" is necessary, please not remove it
 
       - name: Set up MySQL
-        uses: mirromutth/mysql-action@v1.1
+        uses: shogo82148/actions-setup-mysql@v1
```
jfkonecn, line 20: I'll be honest. It makes me a little nervous importing unofficial gitactions.

Is it possible to use docker to run mysql?

calvinlu3: I tried docker and it takes around 24 minutes to do unit tests

jfkonecn (approved):

Issue #3954, opened 2025-09-02
Fix uncurated hotspot mutation being called Oncogenic


calvinlu3:
An uncurated hotspot mutation should not be called Oncogenic and should be capped at Likely Oncogenic

neslihanky:
@calvinlu3 While the fix appears to work nicely for single nucleotide changes, uncurated indel/delins variants can still be annotated as "Oncogenic". In addition, when the variant page is opened, the mutation description shown belongs to a different variant than the one queried. As an example, for the uncurated, made-up variant ERBB2 G776delinsVCAQ, the variant page displays the mutation details of ERBB2 G776delinsLC and annotates the variant as "Oncogenic".

<img width="512" height="74" alt="Image" src="https://github.com/user-attachments/assets/08ddcecb-74eb-4bb0-ac3f-b4aec7e68561" />

<img width="391" height="304" alt="Image" src="https://github.com/user-attachments/assets/046f9224-a3c7-4c17-bb88-d801b5b5f589" />

PR #3958, opened 2025-09-04
Cap oncogenicity of uncurated hotspot mutation at Likely Oncogenic

calvinlu3:
Fixes https://github.com/oncokb/oncokb/issues/3954


bprize15 (approved):
LGTM

Issue #3951, opened 2025-08-29
Wrong annotation for BRAF V600_S602delinsGluLysThr


calvinlu3:
When they search BRAF V600_S602delinsGluLysThr vs BRAF V600_S602delinsEKT (they are the same mutation). It looks like two different alternate alleles are pulled depending on the search.

The correct annotation is `V600_S602delinsEKT`

PR #3959, opened 2025-09-04
fixed alteration annotation for 3 letter aminoacids and log4j

ReshmaRamaiah10:
https://github.com/oncokb/oncokb/issues/3951
- Added a static helper function to convert 3-letter amino acid codes to 1-letter codes before alteration annotation.
Integrated this function in both `Query().setAlteration()` and during `AlterationModel` assignment to ensure consistent handling.
- Updated log4j.properties with Calvin’s suggested changes to improve logging visibility.

### DIFF
--- a/core/src/main/resources/properties-EXAMPLE/log4j.properties
+++ b/core/src/main/resources/properties-EXAMPLE/log4j.properties
@@ -3,21 +3,13 @@
 # -Dlog4j.configuration=file:core/src/main/resources/properties/log4j.properties
 # to setup the file logger.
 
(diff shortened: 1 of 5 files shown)


core/src/main/resources/properties-EXAMPLE/log4j.properties
calvinlu3, line 1: Can you also add some documentation on how we added the log4j file into the Tomcat Server configuration in the README.

calvinlu3 (changes requested):

PR #3960, opened 2025-09-08
Compatibility Matrix

jfkonecn:
One of our users emailed about having compatibility matrix to make local installations easier so I put one together.

Let me know if you see any issues with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,10 +106,15 @@ For this option, you need to download the VEP cache, which is used in the `gn-ve
 
 #### Compatibility Matrix
 
+> [!WARNING]
+> OncoKB uses a native password for MySQL. If you use MySQL 8.1
+> or above you must configure native passwords enabled as they no longer are
+> enabled by default.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,9 +104,17 @@ For this option, you need to download the VEP cache, which is used in the `gn-ve
     docker-compose up -d
     ```
 
+#### Compatibility Matrix
+
+| oncokb | oncokb-transcript | gn-spring-boot | gn-mongo | genome-nexus-vep | MySQL  | Redis  | Oncokb Data |
+| ------ | ----------------- | -------------- | -------- | ---------------- | ------ | ------ | ----------- |
+| v3.x.x | v0.9.4            | v1.x.x         | v0.x     | v0.0.x           | v8.x.x | v7.x.x | v4.x.x      |
```
calvinlu3, line 111: I would limit MySQL to `v8.0.X` because in 8.1, mysql native password was deprecated.
And for genome-nexus-vep, I think we can use `v0.0.1` specifically since there's no other versions after that?

PR #3964, opened 2025-09-25
cap vus oncogencity at likely oncogenic and fix variant description

### LINKED ISSUE #3963: Investigate potential incorrect variant/mutation effect summary for PIK3CA E453_L455del
https://www.oncokb.org/gene/PIK3CA/p.E453_L455del

Think about how categorical alterations are implicitly at most likely oncogenic. Should curation platform make this explicit?

Non-exact matches can at most be likely oncogenic.

We do not want to do partial matching on VUS's. We only want to show last reviewed when it is an exact match. Otherwise, show description similar to https://www.oncokb.org/gene/KRAS/E63L.

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3963

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
@@ -815,7 +815,7 @@ public class IndicatorUtils {
         StringBuilder sb = new StringBuilder("");
         boolean isVus = MainUtils.isVUS(alteration);
         if (isVus) {
-            sb.append(getVUSSummary(getVusDate(alteration), "[[gene]] [[mutation]] [[[mutation]]]", true));
+            sb.append(getVUSSummary("[[gene]] [[mutation]] [[[mutation]]]", getVusDate(alteration), true));
         } else {
             sb.append("The [[gene]] [[mutation]] [[[mutation]]] has not specifically been reviewed by the OncoKB team.");
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/IndicatorUtils.java
```diff
@@ -662,7 +663,7 @@ private static String getMutationEffectDescriptionForUnknownAltButWithAA(Alterat
         StringBuilder sb = new StringBuilder("");
         boolean isVus = MainUtils.isVUS(alteration);
         if (isVus) {
-            sb.append(getVUSSummary(alteration, "[[gene]] [[mutation]] [[[mutation]]]", true));
+            sb.append(getVUSSummary(getVusDate(alteration), "[[gene]] [[mutation]] [[[mutation]]]", true));
```
calvinlu3, line 666: What is the summary suppose to say? This seems to be failing the unit test because it says `last reviewed on gene alteration mutation`

core/src/main/java/org/mskcc/cbio/oncokb/util/SummaryUtils.java
```diff
@@ -818,8 +821,10 @@ public static String hotspotSummary(Alteration alteration, Query query, Boolean
 
     private static String vusAndHotspotSummary(Alteration alteration, Query query, Boolean isHotspot) {
         StringBuilder sb = new StringBuilder();
-        String altStr = getGeneMutationNameInVariantSummary(alteration.getGene(), query.getReferenceGenome(), query.getHugoSymbol(), alteration.getAlteration(), query.isGermline());
-        sb.append(getVUSSummary(alteration, altStr, true));
+        boolean isExactMatch = alteration.getAlteration().equals(query.getAlteration());
```
calvinlu3, line 824: I think the fix should not be in the vusandhostpotsummary function, but more upstream at the findExactlyMatchedAlteration. This function should not use any VUS uinless its an exact match.

calvinlu3: Image

calvinlu3 (changes requested):
Just leaving a comment, so I remember what to fix.

PR #3967, opened 2025-10-03
Updated readme regarding docker hub image location

jfkonecn:


docker-compose.yml
jfkonecn, line 1: Do either one of you have an easy way to test to make sure this file works?

PR #3971, opened 2025-10-07
Default to gene penetrance when variant penetrance is not curated

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/914


bprize15 (approved):
LGTM code-wise, but I think think we should revisit the idea of curating penetrance at both the gene and variant level. Fine to merge to not block the project, but I wonder if we should rethink this.

PR #3978, opened 2025-10-29
Make slight changes to refresh in memory cache

calvinlu3:
Resolves https://github.com/oncokb/oncokb-pipeline/issues/853

### DIFF
--- a/web/src/main/java/org/mskcc/cbio/oncokb/controller/CacheController.java
+++ b/web/src/main/java/org/mskcc/cbio/oncokb/controller/CacheController.java
@@ -102,7 +102,7 @@ public class CacheController {
                     break;
             }
         } catch (Exception e) {
-            LOGGER.error("Cache operation failed for cmd={} (propagation={})", cmd, propagation);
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/mskcc/cbio/oncokb/controller/CacheController.java
```diff
@@ -57,42 +62,51 @@ Map<String, Object> getGeneCache(
...
-        return result;
-    }
-
-    private Boolean resetCache(Boolean propagation) {
-        Boolean operation = true;
-        try {
-            CacheUtils.resetAll(propagation);
         } catch (Exception e) {
-            operation = false;
+            LOGGER.error("Cache operation failed for cmd={} (propagation={})", cmd, propagation);
```
jfkonecn: Include the exception in the log like so:
```java
LOGGER.error("Cache operation failed for cmd={} (propagation={})", cmd, propagation, e);
```

This allows you configure out exception show up via the log4j configuration.

PR #4072, opened 2026-01-21
pick fusion mutation where variant exists over a matched rule

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/1079

Context:
When PAX3 is used as hugo symbol 1, `PAX3-FOXO1 Fusion` is matched: http://localhost:9095/gene/PAX3/somatic/PAX3-FOXO1%20Fusion/RMS

When FOXO1 is used as hugo symbol 1, `Truncating Mutations` is matched: 
https://www.oncokb.org/gene/FOXO1/somatic/Truncating%20Mutations/RMS

This change prioritizes an existing variant instead of picking the first.


calvinlu3 (commented):
Your change makes sense to me - we should prioritize the annotation for the exact match over any of the rules, but I would get a definitive answer from the curators. If they agree, then you can merge the code.

PR #4077, opened 2026-01-28
Add germline annotation endpoints

calvinlu3:
resolve https://github.com/oncokb/oncokb-pipeline/issues/1076


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v2/GermlineAnnotationsApiController.java
```diff
@@ -0,0 +1,793 @@
+package org.mskcc.cbio.oncokb.api.pub.v2;
```
jfkonecn: @bprize15 I need you check this one closely. I don't know the code well enough to be able to catch anything that is not obivous.

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
```diff
@@ -497,6 +495,101 @@ public ResponseEntity<VariantAnnotation> utilVariantAnnotationGet(
...
+        }
+        Alteration alterationModel = null;
+        if (!StringUtils.isNullOrEmpty(hgvsg) || !StringUtils.isNullOrEmpty(genomicChange)) {
+            TranscriptSummaryAlterationResult transcriptSummaryAlterationResult = null;
+            if (!StringUtils.isNullOrEmpty(hgvsg)) {
+                if (this.cacheFetcher.hgvsgShouldBeAnnotated(hgvsg, matchedRG)) {
+                    transcriptSummaryAlterationResult = AlterationUtils.getAlterationFromGenomeNexus(GNVariantAnnotationType.HGVS_G, matchedRG, hgvsg);
+                }
+            } else {
+                if (this.cacheFetcher.genomicLocationShouldBeAnnotated(GenomeNexusUtils.convertGenomicLocation(genomicChange), matchedRG)) {
```
jfkonecn: `else if`

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
```diff
@@ -497,6 +495,101 @@ public ResponseEntity<VariantAnnotation> utilVariantAnnotationGet(
...
+        }
+        query.setTumorType(tumorType);
+        query.setGermline(true);
+
+        List<EvidenceQueryRes> responses = EvidenceUtils.processRequest(Collections.singletonList(query), new HashSet<>(EvidenceTypeUtils.getAllEvidenceTypes(query.isGermline())), LevelUtils.getPublicLevels(), false, false);
+        GermlineIndicatorQueryResp indicatorQueryResp = IndicatorUtils.processQueryGermline(query, null, false, null, false);
+        GermlineVariantAnnotation annotation = new GermlineVariantAnnotation(indicatorQueryResp);
+        EvidenceQueryRes response = responses.iterator().next();
+
+        for (TumorType uniqueTumorType : response.getEvidences().stream().filter(evidence -> !evidence.getCancerTypes().isEmpty()).map(evidence -> evidence.getCancerTypes()).flatMap(Collection::stream).collect(Collectors.toSet())) {
```
jfkonecn: Are you missing a call to `variantAnnotationTumorType.setEvidences`?

jfkonecn (changes requested):

jfkonecn (changes requested):
see comments

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
calvinlu3: No because I got rid of a somatic only logic, so there is no evidence to set for germline.
``` java
// If oncogenic mutations exists when there is also a mutation with a higher level for same cancer type and treament, remove oncogenic mutations
List<Evidence> filteredEvidences = new ArrayList<>();
```

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v2/GermlineAnnotationsApiController.java
```diff
@@ -0,0 +1,793 @@
...
+        String hgvsg,
+        TranscriptSummaryAlterationResult selectedAnnotatedAlteration,
+        ReferenceGenome referenceGenome,
+        List<Alteration> allAlterations
+    ) throws org.genome_nexus.ApiException {
+        Alteration alteration = AlterationUtils.findAlterationWithGeneticType(
+            referenceGenome,
+            hgvsg,
+            allAlterations,
+            germlineQuery.isGermline()
```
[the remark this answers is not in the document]
calvinlu3: I think its still good to have a germline field in the Query object, so that we can easily debug. However, in the code, I don't think we need to use that parameter anymore since we have seperate methods. The isGermline param can be replaced with a hard coded boolean value depending on the method (somatic/germline)

jfkonecn (approved):

jfkonecn (approved):

core/src/main/java/org/mskcc/cbio/oncokb/util/AnnotationSearchUtils.java
```diff
@@ -892,16 +914,30 @@ public int compare(TypeaheadSearchResp e1, TypeaheadSearchResp e2) {
...
-            //Compare Oncogenicity. Treat YES, LIKELY as the same
-            Oncogenicity o1 = Oncogenicity.getByEffect(e1.getOncogenicity());
-            Oncogenicity o2 = Oncogenicity.getByEffect(e2.getOncogenicity());
-            if (o1 != null && o1.equals(Oncogenicity.LIKELY)) {
-                o1 = Oncogenicity.YES;
-            }
-            if (o2 != null && o2.equals(Oncogenicity.LIKELY)) {
-                o2 = Oncogenicity.YES;
+            Integer result;
+            if ((e1.getPathogenicity() != null && !e1.getPathogenicity().isEmpty())
```
bprize15, line 921: Do germline variants get precedence over somatic variants? Also, why do we treat YES and LIKELY the same?

bprize15 (approved):
One question, otherwise LGTM

core/src/main/java/org/mskcc/cbio/oncokb/util/AnnotationSearchUtils.java
calvinlu3: > Do germline variants get precedence over somatic variants?
I'm going off the assumption that there is no overlap between somatic/germline variants. Like somatic cannot be searched with c. right now. I will swap the if condiiton because we usually show somatic first.

> Also, why do we treat YES and LIKELY the same?
I think its because both icons are the same blue target icon. Not sure, so I copied the same behavior as somatic.

PR #4101, opened 2026-04-24
Add linkout to an alternative variant present in oncokb

ReshmaRamaiah10:
https://github.com/oncokb/oncokb-pipeline/issues/1136

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -51,6 +51,7 @@ import org.mskcc.cbio.oncokb.model.FrameshiftVariant;
 import org.mskcc.cbio.oncokb.model.Gene;
 import org.mskcc.cbio.oncokb.model.InferredMutation;
 import org.mskcc.cbio.oncokb.model.Oncogenicity;
+import org.mskcc.cbio.oncokb.model.Query;
 import org.mskcc.cbio.oncokb.model.ReferenceGenome;
 import org.mskcc.cbio.oncokb.model.SpecialVariant;
 import org.mskcc.cbio.oncokb.model.StructuralAlteration;
@@ -621,6 +622,8 @@ public final class AlterationUtils {
             alteration.setName(hgvsc);
         }
         return alteration;
+    }
+
     public static AlternativeOncoKbVariant findAlternativeOncoKbVariant(String hugoSymbol, String alteration, ReferenceGenome referenceGenome) {
         if (StringUtils.isEmpty(hugoSymbol) || StringUtils.isEmpty(alteration)) {
             return null;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -591,6 +592,51 @@ public static Alteration convertTranscriptConsequenceSummaryToAlteration(Transcr
         return alteration;
     }
 
+    public static AlternativeOncoKbVariant findAlternativeOncoKbVariant(String hugoSymbol, String alteration, ReferenceGenome referenceGenome) {
```
jfkonecn, line 629: I'll let calvin review this since knows this code better

jfkonecn (approved):
Please let Calvin review, but I see no problems.

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -591,6 +592,51 @@ public static Alteration convertTranscriptConsequenceSummaryToAlteration(Transcr
...
+
+            String oncokbTranscript = genomeNexusAnnotation.getAnnotationSummary().getTranscriptConsequenceSummary().getTranscriptId();
+            return new AlternativeOncoKbVariant(
+                foundAlteration,
+                foundAlteration.getGene().getHugoSymbol(),
+                inputVariant,
+                oncokbTranscript
+            );
+        } catch (ApiException e) {
+            return null;
```
calvinlu3, line 672: Lets add a comment here why we ignore. I can see a potential issue where GN intermittently fails and we won't be able to find the alternative alteration.

core/src/main/java/org/mskcc/cbio/oncokb/util/SummaryUtils.java
```diff
@@ -28,6 +31,23 @@ public class SummaryUtils {
     public static final String PATHOGENIC_MUTATIONS_DEFAULT_SUMMARY = "\"Pathogenic Mutations\" includes all variants annotated as pathogenic and likely pathogenic by MSK Department of Molecular Genetics.";
     public static final List<String> specialAlterations = Arrays.asList("mutation", "alteration", "insertion", "deletion", "duplication", "fusion", "deletion", "amplification");
 
+    public static AlternativeOncoKbVariant alternativeOncoKbVariant(Query query, Boolean variantExists) {
```
calvinlu3: Seems out of place to put this function in SummaryUtils - maybe AlterationUtils might work?
What do you think about renaming function to `getAlternativeVariantForQuery`? Seems like you are parsing the query object and then sending off to findAlternativeOncokbVariant?

calvinlu3 (changes requested):

web/src/main/java/org/mskcc/cbio/oncokb/api/pvt/PrivateUtilsApiController.java
```diff
@@ -431,6 +431,11 @@ public ResponseEntity<SomaticVariantAnnotation> utilVariantAnnotationGet(
 
         SomaticVariantAnnotation annotation = new SomaticVariantAnnotation(indicatorQueryResp);
         annotation.setAlteration(alterationModel);
+        annotation.setAlternativeOncoKbVariant(AlterationUtils.getAlternativeVariantForQuery(
+            query,
+            indicatorQueryResp.getVariantExist(),
+            indicatorQueryResp.getHasVariantEvidence()
+        ));
```
calvinlu3: I think you can refactor the code, so we do not need to change `IndicatorQueryRespBase` to have JsonIgnore the `hasVariantEvidence` field. The field is being used for internal processing logic rather than part of the api response object.

I thiunk you can just simplify AlterationUtils.getAlternativeVariantForQuery to take just `indicatorQueryResp` because that object already has `indicatorQueryResp.getQuery`. Then you can move the logic from IndicatorUtils Line 500-508 into the method, eliminating the need to compute in another function and then having to create a field just to transport it to another function.

calvinlu3 (changes requested):
LGTM, just a minor refactoring suggestion.

calvinlu3 (approved):

PR #4102, opened 2026-04-29
Germline api tests

calvinlu3:


api_test/scripts/docker-compose.yaml
```diff
@@ -19,10 +19,9 @@ services:
     environment:
       JAVA_OPTS: |
         -Djdbc.driverClassName=com.mysql.jdbc.Driver
-        -Djdbc.url=jdbc:mysql://mysql:3306/oncokb_v7_0_demo?useUnicode=yes&characterEncoding=UTF-8
+        -Djdbc.url=jdbc:mysql://mysql:3306/oncokb_v7_1_demo?useUnicode=yes&characterEncoding=UTF-8
         -Djdbc.username=root
         -Djdbc.password=${MYSQL_PASSWORD}
-        -Dgenomenexus.api=https://www.genomenexus.org/
```
calvinlu3, line 25: This is not an actual config property, so removed.

api_test/golang/queries/genomic_change/2_232006958_232006958_G_T.json
calvinlu3, line 1: The test variant was changed from `2,232006958,232006958,G,T` (DIS3L2, GRCh38) to `7,140710000,140710000,G,T` (NDUFB2, GRCh38) because `genomicLocationShouldBeAnnotated` checks if we have chr2. It builds the gene to chromosome map from `CacheUtils.getAllGenes()`, which in the demo database only contains a small set of genes (BRAF (chr7) , TP53 (chr17), ROS1 (chr6) ). We don't get chr2 back, so we don't make request to GN and so we can't resolve a gene.

A little confused why demo.oncokb.org is sending to GN when it only has 3 genes. I tested locally on master branch and it is not working either. @bprize15

jfkonecn (approved):

api_test/golang/queries/genomic_change/2_232006958_232006958_G_T.json
bprize15: This makes sense, fine with the change.

The reason this is happening is because demo is not connected to transcript. It has no oncokb transcript token. The reason it failed locally for you is most likely because the test used your application.properties which had a token. I have a warning message when the test starts that reminds you to remove your existing properties, but maybe it flashes by too fast. 

The behavior can be explained by: https://github.com/oncokb/oncokb/blob/028272db0430febd70863e4d5234c3987ca218e5/core/src/main/java/org/mskcc/cbio/oncokb/cache/CacheFetcher.java#L336

I think your change makes the test less brittle, so thanks for making it.

calvinlu3: Ohhh, I did not notice that at all. Thanks!

PR #4104, opened 2026-05-06
Fixed exception on big number

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1196

We are getting datadog alerts because someone is querying with a large mutation number. I added a try/catch to deal with.

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
@@ -57,16 +57,16 @@ public class ProteinChangeParser {
 
         result = parseSplice(proteinChange);
         if (result.isParsed) return result;
-        
+
         result = parseRange(proteinChange);
         if (result.isParsed) return result;
-        
+
         result = parseFrameshift(proteinChange);
         if (result.isParsed) return result;
 
         result = parsePoint(proteinChange);
         if (result.isParsed)  return result;
-        
+
         result = parseExtension(proteinChange);
         if (result.isParsed) return result;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
```diff
@@ -332,7 +332,12 @@ public static ParseAlterationResult parseGeneral(String proteinChange) {
             result.normalizedProteinChange = proteinChange.toUpperCase();
 
             ref = m.group(1) == null ? "" : m.group(1).toUpperCase();
-            start = Integer.valueOf(m.group(2));
+            try {
+                start = Integer.valueOf(m.group(2));
+            } catch (NumberFormatException e) {
+                result.setIsParsed(false);
+                return result;
```
jfkonecn, line 339: Seems like returning early is the best way to handle it. If you search for something like `V60000000000000000000000E` it'll say that the variant isn't curated by the OncoKB team.

core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
```diff
@@ -145,7 +145,7 @@ public static ParseAlterationResult parseSplice(String proteinChange) {
         ParseAlterationResult result = new ParseAlterationResult();
         Matcher m = SPLICE_PATTERN.matcher(proteinChange);
         if (m.matches()) {
```
jfkonecn, line 147: The AI decided to fix the random white space. I felt like it probably was for the best to allow it, but let me know if you disagree.

PR #4105, opened 2026-05-06
Fixed exception on big number

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1196

We are getting datadog alerts because someone is querying with a large mutation number. I added a try/catch to deal with.


core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
```diff
@@ -332,7 +332,12 @@ public static ParseAlterationResult parseGeneral(String proteinChange) {
             result.normalizedProteinChange = proteinChange.toUpperCase();
 
             ref = m.group(1) == null ? "" : m.group(1).toUpperCase();
-            start = Integer.valueOf(m.group(2));
+            try {
+                start = Integer.valueOf(m.group(2));
+            } catch (NumberFormatException e) {
+                result.setIsParsed(false);
+                return result;
+            }
```
jfkonecn: Seems like returning early is the best way to handle it. If you search for something like `V60000000000000000000000E` it'll say that the variant isn't curated by the OncoKB team.

core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
```diff
@@ -145,7 +145,7 @@ public static ParseAlterationResult parseSplice(String proteinChange) {
         ParseAlterationResult result = new ParseAlterationResult();
         Matcher m = SPLICE_PATTERN.matcher(proteinChange);
         if (m.matches()) {
-            
+
```
jfkonecn: The AI decided to fix the random white space. I felt like it probably was for the best to allow it, but let me know if you disagree.

core/src/main/java/org/mskcc/cbio/oncokb/util/parser/ProteinChangeParser.java
bprize15: One thing worth considering might be `V6000000000_600000000000001E`. Do we need this logic elsewhere? If we prefer not to handle that case in this PR, that makes sense to me too. 

One other question I have is do we just want to use BigInteger (if this is a use case we want to support)?

bprize15 (approved):
Just one question, otherwise LGTM

bprize15 (approved):
LGTM

PR #4108, opened 2026-05-18
Make sure query object has resolved germline genomic/hgvsg alterations

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1207

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
@@ -550,19 +550,11 @@ public final class AlterationUtils {
         return alt;
     }
 
-    private static Alteration convertTranscriptSummaryAlterationResultToAlteration(TranscriptSummaryAlterationResult transcriptSummaryAlterationResult) {
-        if (transcriptSummaryAlterationResult != null) {
-            return convertTranscriptConsequenceSummaryToAlteration(transcriptSummaryAlterationResult.getTranscriptConsequenceSummary());
-        } else {
-            return null;
-        }
-    }
-
     public static Alteration convertTranscriptConsequenceSummaryToAlteration(TranscriptConsequenceSummary transcriptConsequenceSummary) {
         if (transcriptConsequenceSummary == null) {
             return null;
         }
-        
+
         Alteration alteration = new Alteration();
         String hugoSymbol = transcriptConsequenceSummary.getHugoGeneSymbol();
         Integer entrezGeneId = StringUtils.isNumeric(transcriptConsequenceSummary.getEntrezGeneId()) ? Integer.parseInt(transcriptConsequenceSummary.getEntrezGeneId()) : null;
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
```diff
@@ -613,24 +640,29 @@ public static TranscriptSummaryAlterationResult getAlterationFromGenomeNexus(GNV
...
-                result.setAlteration(alteration);
-            } else {
-                result.setAlteration(new Alteration());
+            TranscriptConsequenceSummary summary = result.getTranscriptConsequenceSummary();
+            Alteration hgvspAlteration = convertTranscriptConsequenceSummaryToAlteration(summary);
+            if (hgvspAlteration != null) {
+                hgvspAlteration.setForGermline(isGermline);
+            }
+            result.setHgvspAlteration(hgvspAlteration != null ? hgvspAlteration : new Alteration());
+            Alteration hgvscAlteration = convertTranscriptConsequenceSummaryToCdnaAlteration(summary);
```
bprize15, line 661: Nit: we could probably parse the both the hgvsc and hgvsp alteration in one pass rather than having to call two functions

ReshmaRamaiah10 (approved):

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/GermlineAnnotationsApiController.java
```diff
@@ -428,7 +430,7 @@ private GermlineIndicatorQueryResp getIndicatorQueryForCuratedHgvs(
     ) throws org.genome_nexus.ApiException {
         Alteration alteration = AlterationUtils.findAlterationWithGeneticType(
             referenceGenome,
-            selectedAnnotatedAlteration.getAlteration().getGene(),
+            selectedAnnotatedAlteration.getHgvscAlteration().getGene(),
```
bprize15, line 428: This assumes that for any curated hgvsg we must have a cDNA change coming from genome nexus right? What happens if the c. is missing?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/GermlineAnnotationsApiController.java
```diff
@@ -711,7 +706,7 @@ private GermlineIndicatorQueryResp getIndicatorQueryFromGenomicLocation(
         Boolean germline,
         Set<EvidenceType> evidenceTypes
     ) {
-        Query query = QueryUtils.getQueryFromAlteration(referenceGenome, tumorType, transcriptSummaryAlterationResult, hgvs);
+        Query query = QueryUtils.getQueryFromAlteration(referenceGenome, tumorType, transcriptSummaryAlterationResult, hgvs, Boolean.TRUE.equals(germline));
```
bprize15: Can we just use `germline` as the last param?

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/GermlineAnnotationsApiController.java
```diff
@@ -749,7 +744,7 @@ private GermlineIndicatorQueryResp getIndicatorQueryFromHGVS(
...
```
bprize15: Same as above

bprize15 (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/util/AlterationUtils.java
calvinlu3: Originally I had one function, but it looked a little bit messy. With one function, I had to return two alteration objects as either a Pair or create my own wrapper class. There are also some hgvsc specific parsing logic like set alteration and name to `hgvsc`, set proteinChange to `hgvsp_short`, and don't set proteinStart and proteinEnd for hgvsc.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/GermlineAnnotationsApiController.java
calvinlu3: I think I put this here because of potential null pointer. But I am making a change, so that it can never be null, which will get rid of this. Since we're in `GermlineAnnotationsAPIController`, I can hardcode query.germline to true, likewise for somatic controller.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/GermlineAnnotationsApiController.java
calvinlu3: If c. is missing from GN response, then hgvscAlteration will not have an `alteration` set, but GN should give us the hugo symbol, which is what we need here. I just need the resolved hugoSymbol from hgvsg/genomic location to find the right gene-alteration. I agree that it not clear because I am calling getHgvscAlteration (technically I can call getHGvspAlteration as well). 

Maybe it can be more clear if the `TranscriptSummaryAlterationResult` has a field `Gene`? Then I can do `selectedAnnotatedAlteration.getGene()`.

bprize15 (approved):
LGTM

PR #4124, opened 2026-06-24
Implement automatic release notes grouping

bprize15:


.github/scripts/check-release-notes.sh
```diff
@@ -12,11 +12,25 @@ if printf '%s\n' "$labels" | grep -Fxq "$skip_label"; then
...
+    echo "✓ ${filename}"
+  done <<< "$release_note_files"
+
   exit 0
 fi
 
 echo "PRs must include a markdown file under release-notes/ unless the '$skip_label' label is applied."
-echo "Expected: add a file like release-notes/pr-123-short-summary.md"
-exit 1
+echo "Expected: add a file like release-notes/pr-123-feat-add-dark-mode.md"
```
jfkonecn, line 35: dark mode is a good feature.

jfkonecn (approved):

.github/scripts/check-release-notes.sh
bprize15: AI seemed to think so. At first I was like why did it change that, but I remembered I asked it to enforce that feat/fix/chore naming

PR #4129, opened 2026-07-09
Fix annotation going through even without alteration

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1241

Given a genomic change for CDKN2A, we can only annotate the p16 (default) isoform. GN gives us transcript ids for both p14 and p16 isoforms.

The root cause is that if there are no matching canonical transcripts, our code uses the most severe consequence and passes it along to processQuery. There seems to be some code that can annotate with just the consequence (mapping truncating consequence terms to Truncating Mutations).


bprize15 (approved):
LGTM

PR #4132, opened 2026-07-16
Resolve hugo symbol to ensembl transcript id before annotating with GN

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1240


bprize15 (approved):
LGTM

PR #4133, opened 2026-07-20
Add inheritance mechanism filter to germline hgvsc endpoint and add gernomic indicator endpoint

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1246

### DIFF
--- a/core/src/main/java/org/mskcc/cbio/oncokb/Constants.java
+++ b/core/src/main/java/org/mskcc/cbio/oncokb/Constants.java
@@ -22,7 +22,7 @@ public final class Constants {
     public static final String PROTEIN_ALTERING_VARIANT = "protein_altering_variant";
 
     public static final String PUBLIC_API_VERSION = "v1.7.0";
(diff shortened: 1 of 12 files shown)


jfkonecn (approved):

core/src/main/java/org/mskcc/cbio/oncokb/Constants.java
```diff
@@ -21,8 +21,8 @@ public final class Constants {
     public static final String UPSTREAM_GENE = "upstream_gene_variant";
     public static final String PROTEIN_ALTERING_VARIANT = "protein_altering_variant";
 
-    public static final String PUBLIC_API_VERSION = "v1.6.0";
-    public static final String PRIVATE_API_VERSION = "v1.6.0";
+    public static final String PUBLIC_API_VERSION = "v1.7.0";
+    public static final String PRIVATE_API_VERSION = "v1.7.0";
```
calvinlu3, line 25: @jfkonecn want to make sure you agree with this change I'm making.

PR #4131, opened 2026-07-15
Validate protein change before annotating

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1242

PR #4137, opened 2026-08-07
update curation validation to get protein sequence in parallel

ReshmaRamaiah10:


calvinlu3 (changes requested):
I think there a less complicated option now since I created a new endpoint in oncokb-transcript that gets you the canonical protein sequence only without the extra bloat that the getCanonicalSequence endpoint currently returns. THat new endpoint is also way faster, so you should not need the complexity from parallelization.
See my pr (https://github.com/oncokb/oncokb/pull/4131/changes#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8).
You should copy my pom.xml changes for the new java client api method and just change to use that endpoint.

calvinlu3:
Adding skip release notes because this is not something that our public or local installation users need to know.

PR #4145, opened 2026-09-03
Name fusions with the HGVS :: separator

calvinlu3:
## Summary

OncoKB curates and reports fusions with the HGVS `::` separator (`BCR::ABL1 Fusion`). A fusion queried with the legacy hyphen is still annotated, and is normalized to the `::` form.

- A fusion queried with a single hyphen (`BCR-ABL1 Fusion`) is rewritten to `BCR::ABL1 Fusion` before anything is looked up, and annotated exactly as the `::` spelling is. The normalized name is what comes back in `query.alteration` and in the generated summaries.
- HUGO symbols may themselves contain hyphens, so a name with more than one hyphen (`H1-4-H2BC5 Fusion`) is resolved by trying every hyphen as the separator: when exactly one split leaves a known gene symbol on both sides, the name is rewritten the same way. Both sides are checked against every gene OncoKB knows of, curated or not, through oncokb-transcript — fusion partners are often uncurated symbols (lncRNAs, IG loci, most HLA genes).
- When no split works, or more than one does, the partners cannot be told apart: the query is reported as a new `AMBIGUOUS_FUSION_SEPARATOR` validation error and is annotated at the gene level only. Written with `::`, the same query is annotated normally.
- A name that is a single hyphenated gene symbol (`COX10-AS1 Fusion`, `HLA-DRB1 Fusion`) has nothing to separate and is left as queried.
- The private `/utils/variantAnnotation` endpoint reports an ambiguous name through its validation object, as `INVALID` with message type `AMBIGUOUS_FUSION_SEPARATOR`. That object is renamed from `proteinChangeValidation` to `variantValidation`, since it now covers fusions, which are not protein changes. A rewritten hyphen is not reported — the annotation is complete.
- Typeahead and annotation search accept both spellings: a keyword typed with a single hyphen is searched for under both.

## API changes

| Parameter/Field Path | Change | Endpoints |
| --- | --- | --- |
| `errors[].type` value `AMBIGUOUS_FUSION_SEPARATOR` | Added | `/annotate/mutations/byProteinChange` (GET, POST), `/annotate/mutations/byHGVSg`, `/annotate/mutations/byHGVSc`, `/annotate/mutations/byGenomicChange`, `/annotate/samples`, `/search` |
| `query.alteration` for a hyphenated fusion query | Edit | Every somatic annotation endpoint |
| `proteinChangeValidation` renamed to `variantValidation` | Edit | `/utils/variantAnnotation` |
| `variantValidation.messageType` value `AMBIGUOUS_FUSION_SEPARATOR` | Added | `/utils/variantAnnotation` |

## Impact

Clients sending fusions with a single hyphen are unaffected in what they get back, except that the fusion name echoed in `query.alteration` and rendered in summary text now uses `::`. Any client comparing those strings against its own hyphenated names needs updating.

A fusion name with more than one hyphen is annotated as before whenever its partners can be told apart; only a name where they cannot receives a validation error instead of an annotation.

`/utils/variantAnnotation` is private and consumed only by oncokb.org, so renaming `proteinChangeValidation` to `variantValidation` affects no API clients. The website needs to read the new field name.

## Deployment note

⚠️ This depends on the fusion rename in the OncoKB data and must be deployed with or after the data release containing it. Against data that still uses hyphenated fusion names, fusion queries will not match. Self-hosted deployments must import that data release alongside this version.

## ⚠️ Blocking pre-merge check: four fusions awaiting curator review

Four curated fusions are still named with hyphens in the data, because they are pending manual curator review:

```
COX10-AS1-NRG1 Fusion      PIK3CA-SOX2-OT fusion
NKX2-5-BCL11B Fusion       PIK3CA-KCNMB2-AS1 fusion
```

These are exactly the fusions whose partner symbol itself contains a hyphen (`COX10-AS1`, `SOX2-OT`, `KCNMB2-AS1`, `NKX2-5`), which is why the rename left them alone.

Each of these names now resolves to a single `::` spelling rather than being reported as ambiguous, so they are no longer a special case in the code. They still do not match while the data keeps the hyphenated names: the query is looked up as `COX10-AS1::NRG1 Fusion` and the curated row is still `COX10-AS1-NRG1 Fusion`.

**Do not merge until curator review completes and these four are renamed to the `::` form in the data release that ships with this code.** Verify afterwards that each returns `variantExist=true`.

## Test plan

- New `FusionValidationUtilsTest`; extended `FusionUtilsTest`, `AlterationUtilsTest`, `IndicatorUtilsTest`, `FindRelevantAlterationsTest`. `FusionUtilsTest` and `FusionValidationUtilsTest` cover the candidate-split rule: rewritten when one split gives two known genes, ambiguous otherwise.
- Test resource TSVs updated to the `::` fusion names.
- Docs: `docs/validation-errors.md`, `release-notes/feat-fusion-hgvs-separator.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


core/src/main/java/org/mskcc/cbio/oncokb/apiModels/ValidationError.java
```diff
@@ -14,7 +14,7 @@
  * See {@code docs/validation-errors.md}.
  */
 public class ValidationError {
-    @ApiModelProperty(value = "What is wrong with the query.", allowableValues = "REFERENCE_ALLELE_MISMATCH, POSITION_OUT_OF_RANGE, REVERSED_POSITION_RANGE, MALFORMED_ALTERATION")
+    @ApiModelProperty(value = "What is wrong with the query.", allowableValues = "REFERENCE_ALLELE_MISMATCH, POSITION_OUT_OF_RANGE, REVERSED_POSITION_RANGE, MALFORMED_ALTERATION, AMBIGUOUS_FUSION_SEPARATOR")
```
jfkonecn, line 17: Is there a way to make it so we don't forget about updating this when a new validation is added?

core/src/main/java/org/mskcc/cbio/oncokb/model/FusionSeparatorStatus.java
```diff
@@ -0,0 +1,18 @@
+package org.mskcc.cbio.oncokb.model;
+
+public enum FusionSeparatorStatus {
```
jfkonecn, line 3: Good comments

jfkonecn (changes requested):

core/src/main/java/org/mskcc/cbio/oncokb/apiModels/ValidationError.java
calvinlu3: By default, it will automatically display all the enum values, so we can omit the allowableValues. The other places where we use allowableValues is to restrict to a subset of an enum (ie LevelOfEvidence enum is use for highestSensitiveLevel, but only Level1-4 are allowed).

jfkonecn (approved):

PR #4148, opened 2026-09-14
Fix: Make ValidationError serializable for Redis caching

jfkonecn:
## Summary
This PR fixes Redis caching failures caused by non-serializable validation error objects in API response payloads.

## Problem
With Redis caching enabled, requests that include validation errors can fail during cache write with `java.io.NotSerializableException: org.mskcc.cbio.oncokb.apiModels.ValidationError`.

## Root cause
Cached response object graphs include `List<ValidationError>` (for example in `SomaticIndicatorQueryResp.errors`), but `ValidationError` did not implement `java.io.Serializable`.

## Fix approach
Make `org.mskcc.cbio.oncokb.apiModels.ValidationError` implement `java.io.Serializable` so response objects containing validation errors can be serialized by the existing Redis cache layer.

## Testing (including reproduction coverage)
none

## Risk / Rollout
Low risk and narrow scope (single model class change). No API contract or payload shape changes expected.

## Related issues
Refs oncokb/oncokb-pipeline#1266


redis.yml
jfkonecn, line 1: Added this so that we can test redis locally

calvinlu3 (approved):

PR #4155, opened 2026-09-21
feat: add hugoSymbol filter to allCuratedGenes endpoint

calvinlu3:
## Summary

Adds an optional `hugoSymbol` request parameter to `GET /utils/allCuratedGenes`, so a client that wants one gene no longer has to download and filter the full curated list (1137 rows / 1028 genes on a current database).

The parameter applies to the latest data version only. Only the JSON endpoint changes. `/utils/allCuratedGenes.txt` is untouched.

## Behavior

| Request | Result |
| --- | --- |
| no `hugoSymbol` | 200, full list — unchanged |
| `hugoSymbol=` (empty, blank, whitespace) | 200, full list — treated as absent |
| `hugoSymbol=BRAF` | 200, BRAF's rows |
| `hugoSymbol=braf` | 404 — matching is **case-sensitive** |
| `hugoSymbol=HER2` | 200, resolves to `ERBB2` — gene aliases are accepted |
| `hugoSymbol=BRCA1` | 200, **two** rows — Somatic and Germline |
| `hugoSymbol=BRAFF` | 404, `No curated gene found for hugoSymbol "BRAFF".` |
| `version=v7.5&hugoSymbol=BRCA1` (latest version) | 200, BRCA1's rows |
| `version=v6.0&hugoSymbol=BRCA1` (any earlier version) | 400, `The hugoSymbol parameter is only supported for the latest data version (v7.5).` |
| `version=v6.0` with no symbol | 200, that version's full list — unchanged |

## Notes on the implementation

**`hugoSymbol` is restricted to the latest data version.** The symbol is resolved against *current* curation (`GeneUtils.getGeneByHugoSymbol`, including the alias fallback) before it is matched, so on an older archive a since-renamed symbol resolves to its current name and then fails to match that archive's records — reporting a gene the archive does contain as missing. This is not hypothetical: 41 of the 42 symbols in the v2.0 payload that have been renamed since (`H3F3A`, `FAM175A`, `WHSC1`, `MRE11A`, `PARK2`, the whole `HIST1H*` family, …) returned 404 while sitting in the very payload being filtered. Rather than resolve symbols per data version, the parameter is now rejected with 400 for anything but the latest version; the message names the current version. v3.0 and later happen to use current symbols throughout, but that is a property of the data, not a guarantee. Archived downloads *without* a symbol are unaffected.

`version` may still be passed together with `hugoSymbol` when it names the latest version, which keeps `?version=<latest>&hugoSymbol=X` working for clients that always pin a version explicitly. The comparison ignores a leading `v`, so `v7.5` and `7.5` are both recognized as latest.

**Filtering happens on the cached list, not inside `CacheFetcher`.** Pushing the filter down into `getCuratedGenesAll` would make `@Cacheable` key on the symbol and store a separate cache entry per query. The full list is small enough that filtering it per request is cheaper than fragmenting the cache.

**A single symbol, not a list.** An earlier draft took `List<String>` to support `?hugoSymbol=BRAF,ERBB2`. That was dropped: a long list runs into Tomcat's 8 KB `maxHttpHeaderSize` (roughly 450–1200 symbols depending on the encoding form) and makes the alias fallback in `GeneUtils.getGeneByAlias` — a full scan of the gene set on each miss — run once per queried symbol. One symbol per request avoids both. Happy to revisit if a caller needs batch lookup.

**404 rather than an empty list.** Returning `200 []` for an unresolvable symbol is indistinguishable from "this gene has no curated entries", which makes a typo silently look like a real answer. The error is thrown as `ApiHttpErrorException`, so it renders through the existing `ExceptionTranslator` with the same `ApiHttpError` body shape as the rest of the API. Note this is 404, whereas `GenesApiController` uses 400 for an unknown gene — 404 reads more accurately here, but it is a deliberate divergence from that precedent and worth a second opinion.

## Testing

`mvn compile` passes. Verified by hand against a local Tomcat deployment with a live database (data version v7.5) and working data-repo credentials, so the `version` paths were exercised end to end this time:

- **Restriction** — `hugoSymbol` + `version=v2.0`/`v6.0`/`v7.1`/`v99.9` all return 400 with the message above, including the `H3F3A`+`v2.0` case that previously returned a wrong 404.
- **Latest version** — `hugoSymbol=BRCA1&version=v7.5` returns 2 rows, `HER2&version=v7.5` returns `ERBB2`, `NOTAGENE&version=v7.5` returns 404. `hugoSymbol=ERBB2&version=v7.5` and `hugoSymbol=ERBB2` agree, so the archive path and the live cache path return the same gene.
- **No regression on archived downloads** — with no symbol: v2.0→669, v3.0→683, v4.0→720, v5.0→942, v6.0→980, v7.0→1008, v7.1→1117, v7.5→1137 rows, all 200. A blank or whitespace symbol alongside an old version still returns that version's full list rather than a 400.
- **No regression on the live path** — baseline 1137 rows, `BRCA1`→2, `HER2`→`ERBB2`, surrounding whitespace trimmed, `brca1`→404, `NOTAGENE`→404, empty→full list. The filtered response is byte-identical to the corresponding subset of the unfiltered response.
- **Untouched endpoints** — `/utils/allCuratedGenes.txt` (with and without `version`/`includeEvidence`), `cancerGeneList`, `cancerGeneList.txt`, `allAnnotatedVariants`, `allActionableVariants`, `genes/lookup`, `annotate/mutations/byProteinChange`, `info` all 200.
- **Swagger** — the Public APIs spec lists `hugoSymbol` as an optional string with the latest-version restriction in its description, and documents the new 400 alongside the existing 404. `/utils/allCuratedGenes.txt` correctly does not list the parameter.

Unrelated to this change: `version=7.5` without the `v` prefix 404s, because the data-repo directory is literally `v7.5`. That happens with or without a symbol.

No automated test was added; there is no existing test class covering `UtilsApiController`. Say the word if you'd like one.

## Release notes

`release-notes/pr-4155-feat-filter-curated-gene-list-by-hugo-symbol.md`, written against the questionnaire in `release-notes/AGENTS.md`. No self-hosting impact: no configuration, environment variable, database, migration, or Docker changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -387,11 +390,28 @@ public ResponseEntity<String> utilsCancerGeneListTxtGet(
...
         @ApiParam(value = VERSION) @RequestParam(value = "version", required = false) String version
         , @ApiParam(value = INCLUDE_EVIDENCE, defaultValue = "TRUE") @RequestParam(value = "includeEvidence", required = false, defaultValue = "TRUE") Boolean includeEvidence
-    ) {
+        , @ApiParam(value = HUGO_SYMBOL) @RequestParam(value = "hugoSymbol", required = false) String hugoSymbol
+    ) throws ApiHttpErrorException {
         if (version != null) {
             return getDataDownloadResponseEntity(version, FileName.ALL_CURATED_GENES, FileExtension.JSON);
         }
-        return new ResponseEntity<>(this.cacheFetcher.getCuratedGenesAll(includeEvidence), HttpStatus.OK);
+        return new ResponseEntity<>(filterCuratedGenesByHugoSymbol(this.cacheFetcher.getCuratedGenesAll(includeEvidence), hugoSymbol), HttpStatus.OK);
```
calvinlu3: **hugoSymbol is silently ignored when `version` is supplied.** The `version != null` branch at line 395 returns the archived download before any filtering, so `GET /utils/allCuratedGenes?version=v4.14&hugoSymbol=BRAF` returns 200 with the entire archived list (~1100 rows) while the caller believes it asked for BRAF only. Worth either applying the filter to the downloaded payload or rejecting the combination with a 400 so the mismatch is not silent.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -387,11 +390,28 @@ public ResponseEntity<String> utilsCancerGeneListTxtGet(
...
+    }
+
+    private List<CuratedGene> filterCuratedGenesByHugoSymbol(List<CuratedGene> genes, String hugoSymbol) throws ApiHttpErrorException {
+        if (StringUtils.isEmpty(hugoSymbol)) {
+            return genes;
+        }
+        String trimmed = hugoSymbol.trim();
+        Gene gene = GeneUtils.getGeneByHugoSymbol(trimmed);
```
calvinlu3: **Alias resolution is case-sensitive, so the documented case-insensitivity does not hold for aliases.** `CacheUtils.getGeneByHugoSymbol` reads an exact-case `HashMap` key, and `GeneUtils.getGeneByAlias` does `gene.getGeneAliases().contains(geneAlias)` — also exact case. So `hugoSymbol=her2` misses both, falls back to `queried = "her2"`, matches no curated symbol and 404s, while `hugoSymbol=HER2` resolves to ERBB2. Plain symbols such as `braf` still work only because of the `equalsIgnoreCase` at line 409, which makes the behavior inconsistent between symbols and aliases. A case-insensitive alias comparison (or an uppercased retry before the alias lookup) would make it match the documented behavior.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -387,11 +390,28 @@ public ResponseEntity<String> utilsCancerGeneListTxtGet(
...
+    ) throws ApiHttpErrorException {
         if (version != null) {
             return getDataDownloadResponseEntity(version, FileName.ALL_CURATED_GENES, FileExtension.JSON);
         }
-        return new ResponseEntity<>(this.cacheFetcher.getCuratedGenesAll(includeEvidence), HttpStatus.OK);
+        return new ResponseEntity<>(filterCuratedGenesByHugoSymbol(this.cacheFetcher.getCuratedGenesAll(includeEvidence), hugoSymbol), HttpStatus.OK);
+    }
+
+    private List<CuratedGene> filterCuratedGenesByHugoSymbol(List<CuratedGene> genes, String hugoSymbol) throws ApiHttpErrorException {
+        if (StringUtils.isEmpty(hugoSymbol)) {
```
calvinlu3: **Whitespace-only `hugoSymbol` 404s instead of being treated as absent.** `StringUtils.isEmpty` runs before the trim, so `?hugoSymbol=%20` (e.g. an empty form field that still submits a space) passes the guard, trims to `""`, matches nothing and returns `404 No curated gene found for hugoSymbol "".`, whereas `?hugoSymbol=` returns the full list. `isBlank` covers both.

```suggestion
        if (StringUtils.isBlank(hugoSymbol)) {
```

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
calvinlu3: Resolved in fe609f1, though not the way this suggests. Rather than making alias lookup case-insensitive, the canonical-symbol path is now case-sensitive too (`equals` instead of `equalsIgnoreCase`), so both paths agree: `BRAF` and `HER2` match, `braf` and `her2` return 404. Case-sensitive matching is the intended contract per the endpoint owner, and it avoids normalizing symbols in a way that would break mixed-case curated entries like `CDKN2A (p14)` and `Other Biomarkers`. The `@ApiParam` description now states that the symbol is case-sensitive.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
calvinlu3: Fixed in fe609f1 — switched to `StringUtils.isBlank`. A whitespace-only value is now treated as absent and returns the full list, consistent with an empty value.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
calvinlu3: Acknowledged, leaving as-is for now. The `version` branch serves a pre-built archived file from GitHub and never materializes the list server-side, so filtering it would mean parsing the archive on every request. The behavior is now documented in the release note ("When `version` is supplied ... `hugoSymbol` is ignored. Filter client-side"). Happy to make the combination a 400 instead if reviewers prefer failing loudly over ignoring silently.

web/src/main/java/org/mskcc/cbio/oncokb/api/pub/v1/UtilsApiController.java
```diff
@@ -387,11 +390,47 @@ public ResponseEntity<String> utilsCancerGeneListTxtGet(
     public ResponseEntity<List<CuratedGene>> utilsAllCuratedGenesGet(
         @ApiParam(value = VERSION) @RequestParam(value = "version", required = false) String version
         , @ApiParam(value = INCLUDE_EVIDENCE, defaultValue = "TRUE") @RequestParam(value = "includeEvidence", required = false, defaultValue = "TRUE") Boolean includeEvidence
-    ) {
+        , @ApiParam(value = HUGO_SYMBOL) @RequestParam(value = "hugoSymbol", required = false) String hugoSymbol
+    ) throws ApiHttpErrorException {
         if (version != null) {
-            return getDataDownloadResponseEntity(version, FileName.ALL_CURATED_GENES, FileExtension.JSON);
+            ResponseEntity<List<CuratedGene>> archived = getDataDownloadResponseEntity(version, FileName.ALL_CURATED_GENES, FileExtension.JSON);
```
calvinlu3: Addressed in 3aeb9b8 — the filter now applies to archived versions rather than being ignored.

Implementation note: `JsonUtils.jsonToArray` uses a `TypeReference<List<T>>` that erases, so the archived body holds `LinkedHashMap`s at runtime even though it is declared `List<CuratedGene>` (the existing cast in `getDataDownloadResponseEntity` is unchecked). A first attempt threw `ClassCastException` on every versioned request — not only from calling `getHugoSymbol()`, but because a generic-typed stream makes javac emit a `checkcast` on the lambda parameter. The filter now reads the symbol through a helper accepting either shape and iterates a raw `List`.

Side benefit: a symbol unknown to the current gene cache still matches the archived payload, so a gene curated under a since-retired symbol stays reachable for the version that contains it.

Caveat: I could not exercise this end to end — my local instance has no GitHub data-repo credentials, so `version=` returns 503 regardless. I verified the filtering directly against `LinkedHashMap` rows from `JsonUtils.jsonToArray`. Please confirm against an environment with data-repo access.

jfkonecn (approved):

